PR #5, opened 2017-07-31
week1.1

massoodbarsa:
Amir week1.1

### DIFF
--- a/examples/basic_mysql_connection/index.js
+++ b/examples/basic_mysql_connection/index.js
@@ -12,12 +12,12 @@ var connection = mysql.createConnection({
 });
 
 connection.connect();
-let query=['SELECT COUNT(*) as items FROM todos',
-           'SELECT * FROM todos WHERE done',
-           'SELECT * FROM todos WHERE !done',
-					 'SELECT * FROM todos GROUP BY DUE DESC',
-					 'SELECT * FROM todos  ORDER BY id DESC LIMIT 1',
-					 'SELECT * FROM todos WHERE Name LIKE "%databases%"'
+let query=['INSERT INTO todos(id,done) VALUES (18,0)',
+           'UPDATE todos SET name="ali",due="2019-08-31" WHERE id=5',
+           'DELETE FROM todos WHERE NAME="ali"',
+					 'UPDATE todos SET done=1 where id=9',
+					 'SELECT * FROM todos'
+
 				  ]
 let sql = query.map(function(i) {
   connection.query(i, function(error, results, fields) {
--- a/examples/basic_mysql_connection/index.js
+++ b/examples/basic_mysql_connection/index.js
@@ -12,7 +12,7 @@ var connection = mysql.createConnection({
 });
 
 connection.connect();
-let query=['INSERT INTO todos(id,done) VALUES '+mysql.escape(18,0),
+let query=['INSERT INTO todos(id,done) VALUES ('+mysql.escape(18,0)),
            'UPDATE todos SET name="ali",due="2019-08-31" WHERE id='+mysql.escape(5),
            'DELETE FROM todos WHERE NAME='+mysql.escape("ali"),
 					 'UPDATE todos SET done=1 where id='+mysql.escape(9),
--- a/examples/basic_mysql_connection/index.js
+++ b/examples/basic_mysql_connection/index.js
@@ -17,7 +17,7 @@ let query=['select count(*) as items from todos',
            'select *  from todos where !done',
 					 'select *  from todos order by due DESC',
 					 'select * from todos where due=(select max(due) from todos)',
-					 'select * from todos where name like "%homework%"'
+					 'select * from todos where name like "%databases%"'
 				  ]
 let sql = query.map(function(i) {
   connection.query(i, function(error, results, fields) {
--- a/examples/basic_mysql_connection/index.js
+++ b/examples/basic_mysql_connection/index.js
@@ -13,9 +13,9 @@ var connection = mysql.createConnection({
 
 connection.connect();
 let query=['select count(*) as items from todos',
-           'select *  from todos where done',
-           'select *  from todos where !done',
-					 'select *  from todos order by due DESC',
+           'select * from todos where done',
+           'select * from todos where !done',
+					 'select * from todos order by due DESC',
 					 'select * from todos where due=(select max(due) from todos)',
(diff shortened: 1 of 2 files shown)


examples/basic_mysql_connection/index.js
```diff
@@ -8,13 +8,22 @@ var connection = mysql.createConnection({
...
 	password: config.password,
 	port: config.port,
-    database: config.database
+  database: config.database
 });
 
 connection.connect();
-
-connection.query('SELECT * FROM todos', function (error, results, fields) {
+let query=['select count(*) as items from todos',
```
jasongwartz: We talked about style issues in class - it will absolutely execute the same, but using capital letters for SQL keywords is a convention that dramatically helps with readability. For next time, can you use `SELECT COUNT(*) AS items FROM todos` instead?

examples/basic_mysql_connection/index.js
```diff
@@ -8,13 +8,22 @@ var connection = mysql.createConnection({
...
+           'select * from todos where done',
+           'select * from todos where !done',
+					 'select * from todos order by due DESC',
```
jasongwartz: Here's another interesting case where case-sensitivity comes into play. Notice:

```
mysql> DESCRIBE todos;
+----------+--------------+------+-----+---------+-------+
| Field    | Type         | Null | Key | Default | Extra |
+----------+--------------+------+-----+---------+-------+
| Id       | int(11)      | NO   |     | 0       |       |
| Done     | tinyint(1)   | YES  |     | 0       |       |
| Name     | varchar(255) | NO   |     | NULL    |       |
| StatusId | int(11)      | YES  |     | NULL    |       |
| Due      | datetime     | YES  |     | NULL    |       |
+----------+--------------+------+-----+---------+-------+
5 rows in set (0.00 sec)
```

The column name is `Due`, not `due`. MySQL doesn't care, but someone who's reading your code might (if they're scanning through your code looking for the place where you make a database lookup for that column).

examples/basic_mysql_connection/index.js
```diff
@@ -8,13 +8,22 @@ var connection = mysql.createConnection({
...
+					 'select * from todos where due=(select max(due) from todos)',
```
jasongwartz: This is an interesting use of a subquery, and not necessarily how I would choose to write this. The question isn't super clear, so I apologise for that.

And, beyond that, this doesn't actually answer the question "Get the single most recently added todo item". Instead, it finds the one that's due the furthest in the future (which is not necessarily the same thing) 😛

examples/basic_mysql_connection/index.js
jasongwartz: Forgot to mention - this doesn't actually accomplish what the question asked for (though it wasn't super clear). I was looking for all todo items sorted with the most recently added ones first.

examples/basic_mysql_connection/index.js
```diff
@@ -12,12 +12,12 @@ var connection = mysql.createConnection({
...
-let query=['select count(*) as items from todos',
-           'select * from todos where done',
-           'select * from todos where !done',
-					 'select * from todos order by due DESC',
-					 'select * from todos where due=(select max(due) from todos)',
-					 'select * from todos where name like "%databases%"'
+let query=['SELECT COUNT(*) as items FROM todos',
+           'SELECT * FROM todos WHERE done',
+           'SELECT * FROM todos WHERE !done',
+					 'SELECT * FROM todos GROUP BY DUE DESC',
```
jasongwartz: Did you mean to use `ORDER BY`? Also, this will retrieve them ordered by when they're due, but not when they're added like the question asked.

jasongwartz:
The new changes are much better! Just one more thing to fix up.

examples/basic_mysql_connection/index.js
```diff
@@ -12,12 +12,12 @@ var connection = mysql.createConnection({
 });
 
 connection.connect();
-let query=['SELECT COUNT(*) as items FROM todos',
-           'SELECT * FROM todos WHERE done',
-           'SELECT * FROM todos WHERE !done',
-					 'SELECT * FROM todos GROUP BY DUE DESC',
-					 'SELECT * FROM todos  ORDER BY id DESC LIMIT 1',
-					 'SELECT * FROM todos WHERE Name LIKE "%databases%"'
+let query=['INSERT INTO todos(id,done) VALUES (18,0)',
```
jalal75: hello dear
is it better for  you try to use mysql.escape() to secure you input

examples/basic_mysql_connection/index.js
```diff
@@ -8,13 +8,22 @@ var connection = mysql.createConnection({
 	user: config.user,
 	password: config.password,
 	port: config.port,
-    database: config.database
+  database: config.database
 });
 
 connection.connect();
+let query=['INSERT INTO todos(id,done) VALUES '+mysql.escape(18,0),
```
jasongwartz: Were you able to run this first query? I get the following error:

```
throw err; // Rethrow non-MySQL errors
        ^

TypeError: Cannot read property '0' of undefined
```

I think we talked about this in the Node class, but you should never push and PR code that doesn't run - it's called 'breaking the build'.

examples/basic_mysql_connection/index.js
```diff
@@ -8,13 +8,22 @@ var connection = mysql.createConnection({
...
+           'UPDATE todos SET name="ali",due="2019-08-31" WHERE id='+mysql.escape(5),
```
jasongwartz: I'm glad you went through some examples and practice here, but this doesn't really address the problem that the homework was asking for. The homework asked you to `Write the necessary SQL statements to complete the functionality of your application`, and I don't think these SQL queries will help the functionality of your todo application (for example, they don't use any user input, and they all use hard-coded Ids).

examples/basic_mysql_connection/index.js
```diff
@@ -8,13 +8,22 @@ var connection = mysql.createConnection({
...
+           'DELETE FROM todos WHERE NAME='+mysql.escape("ali"),
```
jasongwartz: What if the todo item had the name "Ali", or "My name is ali" or "My name is Ali"? Will this query address them, or not?

Also, here you use `NAME` for the field which is called `Name` in the database. Earlier you used `name`. You also use `done` and `id` when the database fields are called `Done` and `Id`. I know this will run - it does work. But it could be a bit confusing for a reader.

PR #6, opened 2017-08-01
SQL HomeWork Week One

AwsIdris:
SQL HomeWork Week One


examples/basic_mysql_connection/index.js
```diff
@@ -4,17 +4,28 @@ var mysql = require('mysql');
...
 	user: config.user,
 	password: config.password,
 	port: config.port,
-    database: config.database
+  database: config.database
 });
 
 connection.connect();
+const sqlStatments=['SELECT COUNT(*) as "Total No of Todos" FROM todos',
+					'SELECT * FROM todos WHERE Done=1',
```
jasongwartz: This works if you assume that `Done` can only be zero or one. What's the type of `Done` as shown in `DESCRIBE todos`? Could it be anything else?

examples/basic_mysql_connection/index.js
```diff
@@ -4,17 +4,28 @@ var mysql = require('mysql');
...
+					'SELECT * FROM todos WHERE Done=0',
+					'SELECT * FROM todos GROUP BY Due DESC',
+					'SELECT * FROM todos WHERE Id=(SELECT MAX(Id) FROM todos) ',
```
jasongwartz: You did something very similar to Amir so I'm going to copy my comments to him here, and then add some more for you:

<hr>

This is an interesting use of a subquery, and not necessarily how I would choose to write this. The question isn't super clear, so I apologise for that.

<hr>

You're very close to what I would consider the 'correct' solution (but there is more than one way). Can you try and come up with another query which returns the same result but without using a subquery?

examples/basic_mysql_connection/index.js
```diff
@@ -4,17 +4,28 @@ var mysql = require('mysql');
...
 	port: config.port,
-    database: config.database
+  database: config.database
 });
 
 connection.connect();
+const sqlStatments=['SELECT COUNT(*) as "Total No of Todos" FROM todos',
+					'SELECT * FROM todos WHERE Done=1',
+					'SELECT * FROM todos WHERE Done=0',
+					'SELECT * FROM todos GROUP BY Due DESC',
```
jasongwartz: I don't think this will return what the question asked for: "Get all the todo items, sorted with the most recent first"

In this case I meant "most recently added". Sorry for the confusion, I'll fix it.

AwsIdris: The question is confusing "Get all the todo items, sorted with the most recent first" because most recent may be ( according to Due Date or according to insertion into the table)?;

examples/basic_mysql_connection/index.js
AwsIdris: It is tinyint it can be (-128 to 127 and null), but usualy we define the value and what does it neans.
for exambel if the value =1 it means done and if value =0 means not done.

jasongwartz: Yeah, that's an application-level decision. But application decisions change - for example, I often write code like `if (todo_is_done) {}`, in which case any 'truthy' value (like 1, or 2, or 124, or "yes") will be the same.

So if my application has added other 'truthy' values in as `Done`, your query will fail. I would probably write this using `WHERE NOT Done=0`. That said, this is a bit confusing to read, so it's also not a perfect solution.

examples/basic_mysql_connection/index.js
jasongwartz: Yes, in this case, I meant most recently added to the table.

Most recently due doesn't really make sense, but it's a bit of a confusing language thing - you would say "due furthest in the future" to describe what your query does.

examples/basic_mysql_connection/index.js
```diff
@@ -13,10 +13,10 @@ var connection = mysql.createConnection({
 
 connection.connect();
 const sqlStatments=['SELECT COUNT(*) as "Total No of Todos" FROM todos',
-					'SELECT * FROM todos WHERE Done=1',
-					'SELECT * FROM todos WHERE Done=0',
-					'SELECT * FROM todos GROUP BY Due DESC',
-					'SELECT * FROM todos WHERE Id=(SELECT MAX(Id) FROM todos) ',
+					'SELECT * FROM todos WHERE Done',
+					'SELECT * FROM todos WHERE !Done',
```
jasongwartz: Yeah, this is more like what I would do 🙂

I prefer the `NOT` keyword over the `!`, but both are equivalent (I just find the `NOT` keyword easier to read).

examples/basic_mysql_connection/index.js
```diff
@@ -13,10 +13,10 @@ var connection = mysql.createConnection({
...
 connection.connect();
 const sqlStatments=['SELECT COUNT(*) as "Total No of Todos" FROM todos',
-					'SELECT * FROM todos WHERE Done=1',
-					'SELECT * FROM todos WHERE Done=0',
-					'SELECT * FROM todos GROUP BY Due DESC',
-					'SELECT * FROM todos WHERE Id=(SELECT MAX(Id) FROM todos) ',
+					'SELECT * FROM todos WHERE Done',
+					'SELECT * FROM todos WHERE !Done',	
+					'SELECT * FROM todos GROUP BY Id DESC',
+					'SELECT * FROM todos  ORDER BY id DESC LIMIT 1',
```
jasongwartz: This, using `LIMIT 1`, is a better answer 🙂

examples/basic_mysql_connection/index.js
```diff
@@ -13,10 +13,10 @@ var connection = mysql.createConnection({
 
 connection.connect();
 const sqlStatments=['SELECT COUNT(*) as "Total No of Todos" FROM todos',
-					'SELECT * FROM todos WHERE Done=1',
-					'SELECT * FROM todos WHERE Done=0',
-					'SELECT * FROM todos GROUP BY Due DESC',
-					'SELECT * FROM todos WHERE Id=(SELECT MAX(Id) FROM todos) ',
+					'SELECT * FROM todos WHERE Done',
+					'SELECT * FROM todos WHERE !Done',	
+					'SELECT * FROM todos GROUP BY Id DESC',
```
jasongwartz: This is better - the `GROUP BY` is unnecessary, though. Since we use `Id` as our unique value, we can be certain that no two rows will have the same `Id` so there will be nothing to group on.

AwsIdris: Dear Jason Thank you for your comments, it was very helpfull to look for beter answars.

examples/basic_mysql_connection/index.js
```diff
@@ -2,30 +2,33 @@ var fs = require('fs');
...
-	console.log("=========================================================");
-  });
-
-
-})
-connection.end();
\ No newline at end of file
+  
+   
+const sqlStatments=[
```
Odai-kakhi: Is that ok to use the same name (sqlStatments) ??

examples/basic_mysql_connection/index.js
```diff
@@ -2,30 +2,33 @@ var fs = require('fs');
...
+        `INSERT INTO todos (Done, Name, StatusId, Due) VALUES (1, 'Review SCSS', 1,'2017-08-17 00:00:00')`,
+        `UPDATE todos SET Name='HTML5' WHERE Id=25`,
+        `UPDATE todos SET StatusId=(select Id FROM statuses WHERE Name='Completed') WHERE Id=27`,
+        `DELETE FROM todos WHERE Id=27`,
+        `DELETE FROM todos WHERE Id=28`
```
jalal75: hello dear
is it better for you try to use mysql.escape() to secure you input?

jasongwartz: +1 to Jalal's comment. You should use `mysql.escape()` anywhere that the data would be coming from untrusted user input. Which values in these statements would probably be coming from user input?

examples/basic_mysql_connection/index.js
```diff
@@ -2,19 +2,33 @@ var fs = require('fs');
...
-connection.query('SELECT * FROM todos', function (error, results, fields) {
-    console.log(results);
-});
-
-connection.end();
\ No newline at end of file
+  
+   
+const sqlStatments=[
+        `INSERT INTO todos (Done, Name, StatusId, Due) VALUES (1, 'Review SCSS', 1,'2017-08-17 00:00:00')`,
```
jasongwartz: Great job using the correct capitalisation (thanks for using `Done` instead of `done` or `DONE`)! 😄

examples/basic_mysql_connection/index.js
```diff
@@ -2,19 +2,33 @@ var fs = require('fs');
...
+        `UPDATE todos SET Name='HTML5' WHERE Id=25`,
+        `UPDATE todos SET StatusId=(select Id FROM statuses WHERE Name='Completed') WHERE Id=27`,
```
jasongwartz: This is a very interesting use of a subquery!

examples/basic_mysql_connection/Schools_Data_Base.sql
```diff
@@ -0,0 +1,71 @@
...
+INSERT INTO aws_students(name, birth_date,gender)
+VALUES ('Ahmad','1980-08-01',1),
+	   ('Adel','1992-10-03',1),
+	   ('Said','1987-07-07',1),
+	   ('Jad','1965-04-05',1),
+	   ('Nada','1980-08-01',0),
+	   ('Suad','1992-10-03',0),
+	   ('Said','1987-07-07',1),
+	   ('Mustafa','1965-04-05',1);
+
```
massoodbarsa: good job aws

PR #7, opened 2017-08-02
HomeWork DataBase

Odai-kakhi:


examples/basic_mysql_connection/index.js
```diff
@@ -4,17 +4,27 @@ var mysql = require('mysql');
...
 	port: config.port,
-    database: config.database
+  database: config.database
 });
 
 connection.connect();
+const sqlStatments=['SELECT COUNT(*) as "Total No of Todos" FROM todos',
+					'SELECT * FROM todos WHERE Done',
+					'SELECT * FROM todos WHERE NOT Done',	
+					'SELECT * FROM todos GROUP BY Id DESC',
```
oelsayed101: Hi Odai,
The changes look perfect for me, I would assume that you meant to use ORDER BY here instead of GROUP BY, right?

examples/basic_mysql_connection/index.js
```diff
@@ -4,17 +4,27 @@ var mysql = require('mysql');
...
+					'SELECT * FROM todos  ORDER BY id DESC LIMIT 1',
+					'SELECT * FROM todos WHERE Name like "%databases%"',
```
oelsayed101: Just a very tiny note here, 'like' needs to be 'LIKE' since you are already following the convention of writing all the SQL keywords in upper case.

examples/basic_mysql_connection/index.js
```diff
@@ -4,17 +4,27 @@ var mysql = require('mysql');
...
-    database: config.database
+  database: config.database
 });
 
 connection.connect();
+const sqlStatments=['SELECT COUNT(*) as "Total No of Todos" FROM todos',
+					'SELECT * FROM todos WHERE Done',
+					'SELECT * FROM todos WHERE NOT Done',	
+					'SELECT * FROM todos GROUP BY Id DESC',
+					'SELECT * FROM todos  ORDER BY id DESC LIMIT 1',
```
jasongwartz: Nice use of `LIMIT 1`.

examples/basic_mysql_connection/index.js
```diff
@@ -4,17 +4,27 @@ var mysql = require('mysql');
...
 	password: config.password,
 	port: config.port,
-    database: config.database
+  database: config.database
 });
 
 connection.connect();
+const sqlStatments=['SELECT COUNT(*) as "Total No of Todos" FROM todos',
+					'SELECT * FROM todos WHERE Done',
+					'SELECT * FROM todos WHERE NOT Done',
```
jasongwartz: Nice use of `NOT Done`. Also good is `! Done`, but I like `NOT` instead - I find it more readable.

examples/basic_mysql_connection/index2.js
```diff
@@ -0,0 +1,30 @@
...
+		host: config.host,
+		user: config.user,
+		password: config.password,
+		port: config.port,
+		database: config.database
+});
+
+
+const sqlStatments=[
+				'INSERT INTO todos (Name, Done) VALUES ("Odai",1)',
```
jasongwartz: The homework asked you to `Write the necessary SQL statements to complete the functionality of your application`. These queries use hardcoded values, so they're probably not very useful for an application (yet).

However, you have achieved *some* of the functionality an application like this would need, like setting the status to "in progress". But you have no query for marking the todo as completed!

examples/basic_mysql_connection/index2.js
```diff
@@ -0,0 +1,30 @@
...
+				'UPDATE todos SET Name = "Odai" WHERE Id=20',
+				`UPDATE todos SET StatusId=2  WHERE Name="odai"`, 
+				'DELETE FROM todos WHERE Name="odai"',
```
jasongwartz: Notice that your name on line 17 is "Odai", but here on line 19, you use "odai". I actually had to test this to figure out what it would do, because I wasn't sure - looks like MySQL ignores case-sensitivity even in string comparisons - but this is REALLY not something I would rely on.

If you want to delete "Odai", ask to delete "Odai". Don't ask to delete "odai" and hope that it does the right thing - this time it did, but in many other similar cases in programming, it won't.

examples/basic_mysql_connection/odai.sql
```diff
@@ -0,0 +1,32 @@
+create table home_owner
+(
+    owner_id        SMALLINT UNSIGNED AUTO_INCREMENT PRIMARY KEY,
+    first_name      VARCHAR(10),
+    last_name       VARCHAR(20)
+);
+
+create table email
+(
+    email_id        SMALLINT UNSIGNED PRIMARY KEY,
```
jasongwartz: Why not auto-increment?

examples/basic_mysql_connection/odai.sql
```diff
@@ -0,0 +1,32 @@
...
+    CONSTRAINT email_owner_FK FOREIGN KEY(email_id) REFERENCES email(email_id) ON DELETE CASCADE
+);
+
+insert into home_owner (first_name, last_name) values
+("Angel", "Flop"),
+("Bob", "Hoe"),
+("Sue", "Hoe");
+
+insert into email values
+(1, "angel@hio.com"),
```
jasongwartz: Same here as before - why not auto-increment these?

examples/basic_mysql_connection/odai.sql
```diff
@@ -0,0 +1,32 @@
...
+    email_id        SMALLINT UNSIGNED PRIMARY KEY,
+    email_address   VARCHAR(40)
+);
+
+create table owner_email
+(
+    owner_id        SMALLINT UNSIGNED,
+    email_id        SMALLINT UNSIGNED,
+    PRIMARY KEY emails_PK(owner_id, email_id),
+    CONSTRAINT owner_email_FK FOREIGN KEY(owner_id) REFERENCES    home_owner(owner_id) ON DELETE CASCADE,
```
jasongwartz: Nice job with these! Looks good :)

PR #8, opened 2017-08-03
The home work week 1 & 2 & 3

jalal75:


examples/basic_mysql_connection/index.js
```diff
@@ -13,8 +13,23 @@ var connection = mysql.createConnection({
 
 connection.connect();
 
-connection.query('SELECT * FROM todos', function (error, results, fields) {
-    console.log(results);
-});
+let query=[	'select count(*) as items from todos',
+			'select * from todos where done',
```
jasongwartz: I notice that you used `select *` instead of `SELECT *`. The convention is to use capital letters for sequel keywords - obviously the lowercase will still work, but the capital letters make it more readable, and readability counts 😀

examples/basic_mysql_connection/index.js
```diff
@@ -13,8 +13,23 @@ var connection = mysql.createConnection({
...
+			'select * from todos where !done',
```
jasongwartz: You're right, in that `! Done` works, but I think `NOT Done` is easier to read.

examples/basic_mysql_connection/index.js
```diff
@@ -13,8 +13,23 @@ var connection = mysql.createConnection({
...
+			'select * from todos order by due DESC',
+			'select * from todos where due=(select max(due) from todos)',
```
jasongwartz: You used a subquery here, similar to Aws and Amir - I'll copy for you what I said to them:

<hr>

This is an interesting use of a subquery, and not necessarily how I would choose to write this. The question isn't super clear, so I apologise for that.

And, beyond that, this doesn't actually answer the question "Get the single most recently added todo item". Instead, it finds the one that's due the furthest in the future (which is not necessarily the same thing) 😛

examples/basic_mysql_connection/index.js
```diff
@@ -13,8 +13,23 @@ var connection = mysql.createConnection({
 
 connection.connect();
 
-connection.query('SELECT * FROM todos', function (error, results, fields) {
-    console.log(results);
-});
+let query=[	'select count(*) as items from todos',
+			'select * from todos where done',
+			'select * from todos where !done',
+			'select * from todos order by due DESC',
```
jasongwartz: This will retrieve the todos ordered by when they're due, but not when they're added. Sorry if the question wasn't clear, I'll work on that 😀

examples/basic_mysql_connection/index.js
```diff
@@ -13,15 +13,15 @@ var connection = mysql.createConnection({
...
+var userInput = 'do your task';
+var sql = [
+		'INSERT INTO todos (name,due) VALUES(' + mysql.escape(userInput) + ', "2017-8-8")',
+		'UPDATE todos SET done= 1   WHERE name= ' + mysql.escape(userInput),
+		'UPDATE todos SET StatusId= 3   WHERE name= ' + mysql.escape(userInput),
+		'DELETe FROM todos WHERE name=' + mysql.escape(userInput)
+	];
+
+let query = sql.map(
 	 (i)=> {
```
Odai-kakhi: I prefer if you used (input OR value) here !!

massoodbarsa: leader leader good job

examples/basic_mysql_connection/index.js
```diff
@@ -13,8 +13,23 @@ var connection = mysql.createConnection({
 
 connection.connect();
 
-connection.query('SELECT * FROM todos', function (error, results, fields) {
-    console.log(results);
-});
+var userInput = 'do your task';
+var sql = [
+		'INSERT INTO todos (name,due) VALUES(' + mysql.escape(userInput) + ', "2017-8-8")',
```
jasongwartz: Thank you for doing this in a method that involves user input, and thank you for using `mysql.escape()`! So far, you're the only person who did that.

examples/basic_mysql_connection/index.js
```diff
@@ -13,8 +13,23 @@ var connection = mysql.createConnection({
...
+		'UPDATE todos SET done= 1   WHERE name= ' + mysql.escape(userInput),
+		'UPDATE todos SET StatusId= 3   WHERE name= ' + mysql.escape(userInput),
+		'DELETe FROM todos WHERE name=' + mysql.escape(userInput)
```
jasongwartz: These all look good to me. Great job!

examples/basic_mysql_connection/Jalal_DataBase.sql
```diff
@@ -0,0 +1,32 @@
+
+create table jalal_doctors(
```
AwsIdris: Dear Jalal It is better to use capital letters Mysql reserved worlds such as CREATE TABLE, and small letters to variables names.
and it is better uo use IF NOT EXISTS when crating a table to avoid any error.

PR #18, opened 2017-09-09
Feedback changes and Playlist Updates

rts-rob:
Removed a confusing link from Lesson1.md

Updated list of videos in Lesson2.md and Lesson3.md to use Lynda.com links.


Lesson1.md
```diff
@@ -41,7 +41,6 @@ Before arriving to class on Sunday, please watch all of the videos in [this vide
 - [Official MySQL Documentation](https://dev.mysql.com/doc/refman/5.7/en/)
 - [Official MySQL Tutorial (pretty dense)](https://dev.mysql.com/doc/refman/5.7/en/tutorial.html)
 - [Node MySQL Documentation on Github](https://github.com/mysqljs/mysql)
-- [Types of SQL JOINs](http://www.khankennels.com/blog/index.php/archives/2007/04/20/getting-joins/)
```
jasongwartz: You don't like these visualisations? I've found them super helpful when explaining inner vs. left vs. outer, etc.

rts-rob: Literally every single student found them confusing. Given that we stay pretty basic (mostly INNER and LEFT) I think we can draw the sets ourselves on the butcher blocks.

PR #41, opened 2018-01-13
Added slides for Week 1 lecture.

fversnel:
The lecture attempts to give the students some intuition
for what databases are and do. It doesn't try to be too concerned
with specific definitions, especially definitions related to details
about SQL. I think it is more useful for the students to grasp the overall
concept of databases and why you would use them.


mkruijt:
sorry @fversnel super late

PR #44, opened 2018-02-05
world queries and TodoERD

yousefallaban:


fdiazsmith:
For the diagram: 
be ready to explain why there is a many to many relationship between task and todo list
Extra: how would you add multiple users on a single task or how would you create a group of list? 

Great work on the queries :)

PR #45, opened 2018-02-06
submitting week1 homework

talalalamdar:


fdiazsmith:
The Diagram could use some love. While it is true that one user can have many tasks. how would you keep track of different lists of tasks? 
or group of lists for that matter? 
or assigning multiple users to one task?

On the queries: 
8 and 9 are close.
8. ask for the languages, and you only provice the quantity of spoken languages. 
9. it ask for the languages, not the sum. 
it will not take much to update these queries. See if you can do it. :)

talalalamdar:
Thanks for the feedback Fernando, you are right about the queries, i made a mistake while reading the question, but about the diagram i am not sure what do you mean ?!

PR #46, opened 2018-02-07
add week1 databases homework

chileluk:


fdiazsmith:
Nice work on the diagram, though I am a little lost why we need a task table. maybe you can explain tomorrow.

On the queries: 
4. you have an extra semi-colon that breaks before the limit
8 and 9. while technically you do provide the answers it would be helpful to provide them with a little bit of useful context. like, what countries are attached to those languages :)

PR #47, opened 2018-02-09
second part from the homework week1

jalalalwani:


fdiazsmith:
Hey, do you have a graph or diagram for the db? 

for the queries:
4. while you do answer it. it would be nice to know the area. 
5. while you do answer it. it would be nice to know the population. 
6. same than above, plus it only ask for top 3.
7. while you do answer it. it would be nice to know the countries.

PR #48, opened 2018-02-09
HomeWork

JawharBairakdar:


fdiazsmith:
4. while you do answer it. it would be nice to know the surface area.
5. while you do answer it. maybe this one has too much information. ID and district may not be that useful?
6. while you do answer it. it would be nice to know the population.  
9. while you do answer it. maybe this one has too much information. ID and district may not be that useful?

PR #50, opened 2018-02-10
homework week1, db

naghamfarwati:


fdiazsmith:
On the diagram: 
It could use some more information. Like what if we want to have list of todos. or a list of list of todos? how would we keep track of tags?
on the queries: 
3 while true, it gives back way to much information. be selective about what your database returns
4. the opposite of above. while true. it would be nice to have the area (not strictly necessary)
5. same than above. but with the population. 
6. same than 5
 8 and 9 ??

naghamfarwati:
week2 db homework

PR #58, opened 2018-03-19
Update world-queries.txt

zydan1:
hamza zydan homework week 1

### DIFF
--- a/Week1/homework/world-queries.txt
+++ b/Week1/homework/world-queries.txt
@@ -38,3 +38,627 @@ SELECT city.Name FROM city JOIN country ON city.CountryCode = country.Code WHERE
 | Alkmaar           |
 +-------------------+
 28 rows in set (0.00 sec)
+
+ select Population,Name from city where Name='Rotterdam';
++------------+-----------+
+| Population | Name      |
++------------+-----------+
+|     593321 | Rotterdam |
++------------+-----------+
+1 row in set (0,00 sec)
+
+3
+
+mysql> select Name ,continent from country where continent = 'North America';
++----------------------------------+---------------+
+| Name                             | continent     |
++----------------------------------+---------------+
+| Aruba                            | North America |
+| Anguilla                         | North America |
+| Netherlands Antilles             | North America |
+| Antigua and Barbuda              | North America |
+| Bahamas                          | North America |
+| Belize                           | North America |
+| Bermuda                          | North America |
+| Barbados                         | North America |
+| Canada                           | North America |
+| Costa Rica                       | North America |
+| Cuba                             | North America |
+| Cayman Islands                   | North America |
+| Dominica                         | North America |
+| Dominican Republic               | North America |
+| Guadeloupe                       | North America |
+| Grenada                          | North America |
+| Greenland                        | North America |
+| Guatemala                        | North America |
+| Honduras                         | North America |
+| Haiti                            | North America |
+| Jamaica                          | North America |
+| Saint Kitts and Nevis            | North America |
+| Saint Lucia                      | North America |
+| Mexico                           | North America |
+| Montserrat                       | North America |
+| Martinique                       | North America |
+| Nicaragua                        | North America |
+| Panama                           | North America |
+| Puerto Rico                      | North America |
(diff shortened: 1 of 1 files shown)


Week1/homework/world-queries.txt
```diff
@@ -37,4 +37,628 @@ SELECT city.Name FROM city JOIN country ON city.CountryCode = country.Code WHERE
...
++------------+-----------+
+| Population | Name      |
++------------+-----------+
+|     593321 | Rotterdam |
++------------+-----------+
+1 row in set (0,00 sec)
+
+3
+
+mysql> select Name ,continent from country where continent = 'North America';
```
jasongwartz: Be very careful with your capital letters here. In databases, same as variables when programming, `Continent` may not be the same as `continent`.

Week1/homework/world-queries.txt
```diff
@@ -37,4 +37,628 @@ SELECT city.Name FROM city JOIN country ON city.CountryCode = country.Code WHERE
 | Heerlen           |
 | Alkmaar           |
 +-------------------+
-28 rows in set (0.00 sec)
\ No newline at end of file
+28 rows in set (0.00 sec)
+
+ select Population,Name from city where Name='Rotterdam';
```
jasongwartz: A convention, though not a rule, is to put the SQL keywords in capital letters - like `SELECT Population, Name FROM city`. It makes it easier to identify which words are column or table names and which words are SQL keywords.

jasongwartz: Also, be careful with where you insert spaces. I would suggest `Population, Name` instead of `Population,Name` as you've currently written.

Week1/homework/world-queries.txt
```diff
@@ -37,4 +37,628 @@ SELECT city.Name FROM city JOIN country ON city.CountryCode = country.Code WHERE
...
+|    8389200 | Moscow            |
+|    8008278 | New York          |
++------------+-------------------+
+10 rows in set (0,09 sec)
+
+
+6
+
+
+select Name,Population  from country where Name LIKE 'p%'  ORDER BY Population DESC LIMIT 3 ;
```
jasongwartz: Again, be careful - in this case it worked, but in other cases, there may be a difference between `'P%'` and `'p%'`.

Week1/homework/world-queries.txt
```diff
@@ -37,4 +37,628 @@ SELECT city.Name FROM city JOIN country ON city.CountryCode = country.Code WHERE
...
++-----------------+
+|      3705025700 |
++-----------------+
+1 row in set (0,00 sec)
+
+
+8
+
+
+mysql> select distinct  country.Region , countrylanguage.Language from country,countrylanguage where country.Region = 'South America';
```
jasongwartz: This isn't correct. You've accidentally created a 'cartesian product', where every row in `country` is multiplied by every row in `countrylanguage`. Notice that your query returns the same number of rows as `SELECT DISTINCT Language FROM countrylanguage` - well, every language in the world isn't spoken in South America!

Try and review `JOIN` - remember that you need to specify what they have in common (what they will join `ON`).

Week1/homework/world-queries.txt
```diff
@@ -37,4 +37,628 @@ SELECT city.Name FROM city JOIN country ON city.CountryCode = country.Code WHERE
...
+457 rows in set (0,02 sec)
+
+mysql> 
+
+
+
+9 
+
+
+mysql> select city.Name , city.CountryCode,countrylanguage.Language  from city INNER  JOIN  countrylanguage ON countrylanguage.CountryCode = city.CountryCode where city.Name = 'Barcelona' ;
```
jasongwartz: Yes, this is technically correct, since both tables contain the same shared value from `country` - they both use `CountryCode`. However, if they were both related to `country` but used different values to relate to `country`, this wouldn't work - you would have to do a three-table join.

Week1/homework/world-queries.txt
jamsrahimi: Hi @zydan1 or @fadikurdi, did you fix or comment here yet?

PR #63, opened 2018-03-24
DB-HW1 Done

jamsrahimi:
Hi Jason,
The assignment is Done I guess.


Week1/homework/world-queries.txt
```diff
@@ -141,12 +141,12 @@ SELECT Name AS CountryName FROM Country WHERE Name LIKE 'P%' ORDER BY Population
 
 7. What is the population in Asia?
 
- SELECT Continent, Population FROM Country WHERE Continent='Asia' GROUP BY Continent;
-+-----------+------------+
-| Continent | Population |
-+-----------+------------+
-| Asia      |   22720000 |
-+-----------+------------+
+ SELECT Continent, SUM(Population) FROM Country WHERE Continent='Asia' GROUP BY Continent;
```
jasongwartz: 👍

jasongwartz: Though since we've already filtered for only Asia, you could also have written this:
```
SELECT SUM(Population) FROM country WHERE Continent = 'Asia'
```

Alternatively, your solution would work well if you wanted the population of all continents. Try running this, for example:

```
SELECT Continent, SUM(Population) as totalContinentPopulation, COUNT(*) as numberOfCountries FROM country GROUP BY Continent
```

jamsrahimi: Thanks Jason.

PR #72, opened 2018-04-07
DianaHW2

DianaAyoub:
Diana datatbases homework2

### DIFF
--- a/Week2/MAKEME.md
+++ b/Week2/MAKEME.md
@@ -18,36 +18,110 @@ implement them:
 
 *Read how to [use and escape query values](https://github.com/mysqljs/mysql#escaping-query-values)*
 
-```js
+// This is the connector (also known as driver)
+// that we can use to connect to a MySQL process
+// and access its databases.
+const mysql = require('mysql');
+
 class TodoModel {
-
-    // ...
-
-    create(description, callback) {
-        // Write code and query to create a new TODO item
+    constructor(dbConnection) {
+        this.dbConnection = dbConnection;
     }
 
-    update(id, description, callback) {
-        // Write code and query to update and existing TODO item
+    // Loads all the TODOs in the database
+    load(callback) {
+        const selectTodoItems = "SELECT * FROM todo_items";
+        this.dbConnection.query(selectTodoItems, function (err, results, fields) {
+            if (err) {
+                callback(err);
+                return;
+            }
+
+            callback(null, results);
+        });
     }
 
-    delete(id, callback) {
-        // Write code and query to delete an existing TODO item
-    }
 
-    tagTodoItem(todoItemId, tagId, callback) {
-        // Write code and query add a tag to a TODO item
-    }
-        
-    untagTodoItem(todoItemId, tagId, callback) {
-        // Write code and query remove a tag from a TODO item
-    }
 
-    markCompleted(todoItemId, callback) {
-        // Write code to mark a TODO item as completed
-    }
+
+    function create(id, description, callback) {
+    // Write code and query to create a new TODO item
+    var query = dbConnection.query('INSERT INTO todo_items (text, user_id) VALUES ( ' + description + ' , ' + id + ' )');
+    query.on('Result', function (row) {
+        callback(null, row.id)
+    });
 }
-```
+
+
+
+function update(id, description, callback) {
+    // Write code and query to update and existing TODO item
+    var query = dbConnection.query('UPDATE todo_items SET text = ' + description + 'Where user_id = ' + id);
+    query.on('Result', function (row) {
+        callback(null, row.id)
+    });
+}
+
+
+
+function delete (id, callback) {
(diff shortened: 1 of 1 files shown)


Week2/MAKEME.md
```diff
@@ -18,36 +18,110 @@ implement them:
...
-        // Write code and query remove a tag from a TODO item
-    }
 
-    markCompleted(todoItemId, callback) {
-        // Write code to mark a TODO item as completed
-    }
+
+    function create(id, description, callback) {
+    // Write code and query to create a new TODO item
+    var query = dbConnection.query('INSERT INTO todo_items (text, user_id) VALUES ( ' + description + ' , ' + id + ' )');
```
jasongwartz: Good job here - but I think the sample code in the other file would have helped a lot.

In general, composing a string (in this case, the string that will become the query) in this way isn't the best. You've used 'string concatenation', where you add two strings together with '+'. It makes it a bit difficult to read, and it's also open to SQL injection vulnerabilities.

You could also use 'string templating', where the string is written with 'backticks', the ` character. When you do that, you can write a string like this:
```
db.Connection.query(`INSERT INTO todo_items (text, user_id) VALUES ( ${ description }, ${ id } )
```
String templating like this is still open to a SQL injection vulnerability. To fix that, you can use `connection.escape()` on the strings to escape them, or use a 'prepared statement'.

Week2/MAKEME.md
```diff
@@ -18,36 +18,110 @@ implement them:
...
+        callback(null, row.id)
+    });
 }
-```
+
+
+
+function update(id, description, callback) {
+    // Write code and query to update and existing TODO item
+    var query = dbConnection.query('UPDATE todo_items SET text = ' + description + 'Where user_id = ' + id);
```
jasongwartz: This is not correct -you'd be changing all of the todo_items associated with that user, since you wrote `WHERE user_id = ...`.

Week2/MAKEME.md
```diff
@@ -18,36 +18,110 @@ implement them:
...
-    markCompleted(todoItemId, callback) {
-        // Write code to mark a TODO item as completed
-    }
+
+    function create(id, description, callback) {
+    // Write code and query to create a new TODO item
+    var query = dbConnection.query('INSERT INTO todo_items (text, user_id) VALUES ( ' + description + ' , ' + id + ' )');
+    query.on('Result', function (row) {
+        callback(null, row.id)
+    });
```
jasongwartz: You don't handle what happens if the query errors here. Take a look at the other homework file with sample code to see how it can be done :smile:

Week2/MAKEME.md
```diff
@@ -18,36 +18,110 @@ implement them:
...
+}
+
+
+
+function delete (id, callback) {
+    // Write code and query to delete an existing TODO item
+    var query = dbConnection.query('DELETE FROM todo_items WHERE user_id = ' + id);
```
jasongwartz: Same issue here as with `UPDATE`, you'll delete every todo for the user specified by 'id'. In this case, 'id' should mean the todo_item id, not the user id.

Week2/MAKEME.md
```diff
@@ -18,36 +18,110 @@ implement them:
...
+    query.on('Result', function (row) {
+        callback(null, row.id)
+    });
+}
+
+function tagTodoItem(todoItemId, tagId, callback) {
+    // Write code and query add a tag to a TODO item
+    var query = dbConnection.query('INSERT INTO  todo_item_tag(todoItemId, tagId) VALUES ( ' + todoItemId + ' , ' + tagId + ' )');
```
jasongwartz: :+1: 

This doesn't handle the case where a new tag is being created, though - ie. that tag has not been used before and doesn't already exist.

Week2/MAKEME.md
```diff
@@ -18,36 +18,110 @@ implement them:
...
+    query.on('Result', function (row) {
+        callback(null, row.todoItemId)
+    });
+}
+
+
+function untagTodoItem(todoItemId, tagId, callback) {
+    // Write code and query remove a tag from a TODO item
+    var query = dbConnection.query('DELETE FROM todo_item_tag WHERE todoItemId = ' + todoItemId + 'AND tagId = ' + tagId);
```
jasongwartz: :+1:

Week2/MAKEME.md
```diff
@@ -18,36 +18,110 @@ implement them:
...
+    query.on('Result', function (row) {
+        callback(null, row.todoItemId)
+    });
+}
+
+function markCompleted(todoItemId, callback) {
+    // Write code to mark a TODO item as completed
+    var query = dbConnection.query('UPDATE todo_items SET is_completed = TRUE WHERE id = ' + todoItemId);
```
jasongwartz: Nice job using `TRUE` for the value, a lot of people used `1` (which also works, but yours is better and more clear).

PR #79, opened 2018-04-13
Db week3 mahmut

mahmutkaya:


Week1/homework/world-queries.txt
```diff
@@ -37,4 +37,181 @@ SELECT city.Name FROM city JOIN country ON city.CountryCode = country.Code WHERE
 | Heerlen           |
 | Alkmaar           |
 +-------------------+
-28 rows in set (0.00 sec)
\ No newline at end of file
+28 rows in set (0.00 sec)
```
jasongwartz: Try not to add so much to each PR, since it's not actually relevant. You've included all the previous homework stuff in the week3 PR, which just makes it more difficult for your reviewer to find what's relevant.

Week3/school_dump.sql
```diff
@@ -0,0 +1,247 @@
...
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `parent` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `name` varchar(50) DEFAULT NULL,
+  `relationship` varchar(50) NOT NULL,
+  `address` varchar(200) DEFAULT NULL,
+  `student_id` int(11) NOT NULL,
+  PRIMARY KEY (`id`),
+  CONSTRAINT `parent_ibfk_1` FOREIGN KEY (`id`) REFERENCES `student` (`id`)
```
jasongwartz: This assumes each parent can only be the parent of one student, but what if two or more siblings go to the same school? I think this should be a many-many relationship, but you've modeled it as one-many here. Think - one student can have many parents, but one parent can have many students.

If you think one parent can have many students, but each student can only have one parent, then you can make it a one-many relationship - but in the other direction, where each 'student' record contains a reference to a parent.

Week3/school_dump.sql
```diff
@@ -0,0 +1,247 @@
...
+DROP TABLE IF EXISTS `subject`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `subject` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `name` varchar(50) NOT NULL,
+  `class_id` int(11) NOT NULL,
+  PRIMARY KEY (`id`),
+  KEY `class_id` (`class_id`),
+  CONSTRAINT `subject_ibfk_1` FOREIGN KEY (`class_id`) REFERENCES `class` (`id`)
```
jasongwartz: I think this is incorrect. You've created a one-many relationship here as well, where each subject can only be studied by one class (one class can have many subjects), but that doesn't make sense - class and subject should be many-many.

Week3/school_dump.sql
```diff
@@ -0,0 +1,247 @@
...
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `teacher` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `name` varchar(50) NOT NULL,
+  `address` varchar(200) DEFAULT NULL,
+  `subject_id` int(11) NOT NULL,
+  PRIMARY KEY (`id`),
+  KEY `subject_id` (`subject_id`),
+  CONSTRAINT `teacher_ibfk_1` FOREIGN KEY (`subject_id`) REFERENCES `subject` (`id`)
```
jasongwartz: Same problem here. I am a teacher, and I teach 'Node' but also 'Databases', and this can't be modeled in your table. You'd have to create two entries of me, which doesn't make sense. This should also be many-many.

Week3/school_dump.sql
```diff
@@ -0,0 +1,247 @@
...
+  `subject_id` int(11) NOT NULL,
+  `class_id` int(11) NOT NULL,
+  `teacher_id` int(11) NOT NULL,
+  PRIMARY KEY (`id`),
+  KEY `subject_id` (`subject_id`),
+  KEY `class_id` (`class_id`),
+  KEY `teacher_id` (`teacher_id`),
+  CONSTRAINT `time_table_ibfk_1` FOREIGN KEY (`subject_id`) REFERENCES `subject` (`id`),
+  CONSTRAINT `time_table_ibfk_2` FOREIGN KEY (`class_id`) REFERENCES `class` (`id`),
+  CONSTRAINT `time_table_ibfk_3` FOREIGN KEY (`teacher_id`) REFERENCES `teacher` (`id`)
```
jasongwartz: You were closer to being able to model the many-many relationship here, but still didn't quite get it right.

PR #80, opened 2018-04-14
Jamshid db 3

jamsrahimi:
I hope this is enough.


Week3/Company_create_insert_fk.sql
```diff
@@ -0,0 +1,104 @@
...
+CREATE TABLE IF NOT EXISTS Employee (
+  eID INT(4) NOT NULL,
+  eName VARCHAR(45) NOT NULL,
+  initials VARCHAR(10) NULL DEFAULT NULL,
+  dateOfBirth DATE NOT NULL,
+  position VARCHAR(20) NOT NULL,
+  salary DECIMAL(7,2) NOT NULL,
+  room VARCHAR(10) NOT NULL,
+  bossNr INT(4) NULL DEFAULT NULL,
+  dID INT(2) NOT NULL,
```
jasongwartz: You're not paying by the letter, you can feel free to use 'employeeID' and 'departmentID' instead of 'eID' and 'dID'. :stuck_out_tongue: 

And you should, because they're more clear and more readable.

Week3/Company_create_insert_fk.sql
```diff
@@ -0,0 +1,104 @@
...
+  description VARCHAR(45) NOT NULL,
+  country VARCHAR(45) NOT NULL,
+  mgrID INT(4) NULL DEFAULT NULL,
+  PRIMARY KEY (code));
+
+CREATE TABLE IF NOT EXISTS Department (
+  dID INT(2) NOT NULL,
+  dName VARCHAR(25) NOT NULL,
+  LocationCode VARCHAR(5) NOT NULL,
+  secretaryID INT(4) NULL DEFAULT NULL,
```
jasongwartz: Shouldn't this be a foreign key to an employee?

Week3/Company_create_insert_fk.sql
```diff
@@ -0,0 +1,104 @@
...
+  room VARCHAR(10) NOT NULL,
+  bossNr INT(4) NULL DEFAULT NULL,
+  dID INT(2) NOT NULL,
+  PRIMARY KEY (eID));
+
+CREATE TABLE IF NOT EXISTS Location (
+  code VARCHAR(5) NOT NULL,
+  description VARCHAR(45) NOT NULL,
+  country VARCHAR(45) NOT NULL,
+  mgrID INT(4) NULL DEFAULT NULL,
```
jasongwartz: Shouldn't this be a foreign key to a manager?

Week3/Company_create_insert_fk.sql
jasongwartz: If you don't create a foreign key constraint, then it will be possible to put any integer - whether it's a real employee ID or not - as this value. But because of the 'C' in ACID (consistency), the data must always follow the rules, and we definitely want it to be the case that if there is a value here, it's a real employee.

Week3/Company_create_insert_fk.sql
```diff
@@ -0,0 +1,104 @@
...
+
+CREATE TABLE IF NOT EXISTS SalaryScale (
+  scale INT(2) NOT NULL,
+  lowerLimit DECIMAL(7,0) NOT NULL,
+  upperLimit DECIMAL(7,0) NOT NULL,
+  PRIMARY KEY (scale));
+
+ALTER TABLE Employee
+  ADD CONSTRAINT Employee_has_boss_fk
+    FOREIGN KEY (bossNr) REFERENCES Employee (eID),
```
jasongwartz: Ah, I see - you added the constraints down here.
It would be more clear to make them part of the CREATE TABLE.

PR #81, opened 2018-04-14
Diana databases  homework3 ...

DianaAyoub:
this is my homework


Week3/README.md
```diff
@@ -1,57 +1,82 @@
...
+  INDEX fk_supermarket_purchase 1_idx (purchase_id  ASC),
+  CONSTRAINT fk_supermarket_product1
+  FOREIGN KEY (products_id)
+  REFERENCES supermarket.products (id)
+  ON DELETE NO ACTION
+  ON UPDATE NO ACTION,
+  CONSTRAINT  fk_supermarket_purchase1
+  FOREIGN KEY (purchase_id)
+  REFERENCES supermarket.purchase (id)
+  ON DELETE NO ACTION
```
jasongwartz: We didn't talk about cascades in class - are you certain you know what this does? I think it's probably correct for this table, but its not necessary unless another table is referencing this table, which is probably not the case - this table is referencing others.

PR #82, opened 2018-04-15
database week3 homework added

arasmr:
database week3 homework added


Week3/mybusiness.sql
```diff
@@ -0,0 +1,189 @@
...
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `departments` (
+  `dep_number` int(11) NOT NULL AUTO_INCREMENT,
+  `employee_id` int(11) NOT NULL,
+  `department_name` varchar(50) NOT NULL,
+  PRIMARY KEY (`dep_number`),
+  KEY `employee_id` (`employee_id`),
+  CONSTRAINT `departments_ibfk_1` FOREIGN KEY (`employee_id`) REFERENCES `employees` (`id`)
+) ENGINE=InnoDB AUTO_INCREMENT=2 DEFAULT CHARSET=utf8;
```
jasongwartz: Did you write these by hand or with some kind of tool? I would have preferred to see actual CREATE TABLE statements written by you, not the output of sqldump

Week3/mybusiness.sql
```diff
@@ -0,0 +1,189 @@
...
+--
+-- Table structure for table `departments`
+--
+
+DROP TABLE IF EXISTS `departments`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `departments` (
+  `dep_number` int(11) NOT NULL AUTO_INCREMENT,
+  `employee_id` int(11) NOT NULL,
```
jasongwartz: This is creating a 1 to 1 relationship, which strikes me as strange for employee <-> department. I don't think this is correct

Week3/mybusiness.sql
```diff
@@ -0,0 +1,189 @@
...
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `employees` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `full_name` varchar(50) NOT NULL,
+  `date_of_birth` date NOT NULL,
+  `address` varchar(50) NOT NULL,
+  `phone_number` varchar(50) NOT NULL,
+  `e_mail` varchar(50) NOT NULL,
+  PRIMARY KEY (`id`)
```
jasongwartz: I think you just got this backwards. Each employee could have an attribute 'departmentID', which assigns an employee to one department. This makes a 1 to many relationship in the direction of department -> employee (as in: 1 employee has 1 department, 1 department has many employees). Right now, you've created the opposite - 1 employee has many departments, 1 department has 1 employee - and I don't think that makes sense.

Week3/mybusiness.sql
```diff
@@ -0,0 +1,189 @@
...
+/*!40000 ALTER TABLE `employees` DISABLE KEYS */;
+INSERT INTO `employees` VALUES (1,'emre','1991-06-15','Netherland','12344566','earas446@gmail.com');
+/*!40000 ALTER TABLE `employees` ENABLE KEYS */;
+UNLOCK TABLES;
+
+--
+-- Table structure for table `project_info`
+--
+
+DROP TABLE IF EXISTS `project_info`;
```
jasongwartz: This seems like a table that creates the many-to-many relationship between projects and departments - so, one project can be worked on by many departments, and one department can work on many projects. This makes sense, but then the table name is wrong - it should be `project_department` (would be the normal way of naming it) or something else that makes that more clear.

Week3/mybusiness.sql
```diff
@@ -0,0 +1,189 @@
...
+UNLOCK TABLES;
+
+--
+-- Table structure for table `project_team`
+--
+
+DROP TABLE IF EXISTS `project_team`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `project_team` (
```
jasongwartz: Again, a many-to-many of projects to employees. But that's not really a team, per se - you'll end up with one record per employee per project, and the table name implies that one record would be one team, which is not the case.

Week3/mybusiness.sql
```diff
@@ -0,0 +1,189 @@
...
+UNLOCK TABLES;
+
+--
+-- Table structure for table `salaries`
+--
+
+DROP TABLE IF EXISTS `salaries`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `salaries` (
```
jasongwartz: This might be better as an attribute of the employees table, since it's probably one-to-one.

PR #85, opened 2018-04-20
homework week3 DB

elmira202003:
Dear Jason,
This database is for an online hand painted bag store . thanks in advance for your attention.


Week3/mysql.sql
```diff
@@ -0,0 +1,70 @@
+CREATE database bag;
+use bag;
+
+
+CREATE TABLE `product` ( 
+    `id` int(10) NOT NULL AUTO_INCREMENT, 
+    `name` varchar(45) NOT NULL , 
+    `price` double ,
```
jasongwartz: I think we talked about this in class - but if we didn't, you should read up on why 'double' types, or any floating-point type, is a really bad choice for storing fixed-point numbers like prices.

Week3/mysql.sql
```diff
@@ -0,0 +1,70 @@
...
+     PRIMARY KEY (`id`));
+
+
+
+CREATE tABLE `product material` (
```
jasongwartz: You can't have spaces in table names.

Week3/mysql.sql
```diff
@@ -0,0 +1,70 @@
...
+    `size` varchar(45),
+    `number_of_paint_colors` int(10), 
+    `color_type` varchar(45),
+    `leather_type` varchar(45),
+    UNIQUE (`product_id`), 
+    FOREIGN KEY (`product_id`) REFERENCES `product` (`id`));
+
+
+
+CREATE tABLE `product material` (
```
jasongwartz: I think you put this one in twice?
But you wrote the foreign key differently.

Week3/mysql.sql
```diff
@@ -0,0 +1,70 @@
...
+    `email` varchar(45) NOT NULL ,
+    `password` varchar(45) NOT NULL ,
+    `first_name` varchar(45),
+    `last_name` varchar(45),
+    PRIMARY KEY (`id`),
+    UNIQUE KEY `email` (`email`));
+
+
+
+CREATE TABLE `order` (
```
jasongwartz: This schema is a bit confusing - can a customer order only one item at a time? This schema implies that a customer gets a new order_id for every product_id they purchase. What if they want to order two different things at once, or two of something?

PR #92, opened 2018-05-08
Hadeel homework week1 

hadeel43:


Week1/homework/world-queries.txt
```diff
@@ -37,4 +37,166 @@ SELECT city.Name FROM city JOIN country ON city.CountryCode = country.Code WHERE
...
+| Hindi            |
+| Sranantonga      |
+| Goajiro          |
+| Warrau           |
++------------------+
+21 rows in set (0.00 sec)
+
+9. What are the languages spoken on all cities named ‘Barcelona’ (you may need to join 3 tables =) ) 
+
+mysql> SELECT distinct language FROM countrylanguage JOIN city ON countrylanguage.countrycode=city.countrycode WHERE city.name='Barcelona';
```
pcelentano: Interesting....  `countrylanguage` and `city` are not directly related.. but they have an implicit connection through `countrycode`...  best practice is to join tables that are directly related `countrylanguage` with `country` and that result join it again with `city` (or the other way around) (Ill tell Nadine to explain this one better in class, if she doesn't please ask about this one !)

As a sidenote part of the fun of this query (besides trying to join 3 tables is also realizing there is more than one country with a city named Barcelona) so you can also return country name, city name (will be a column with all values Barcelona) and the language.. then you'll need to remove the distinct ! Give it a try !

If you need the entity-relation diagram 
http://4.bp.blogspot.com/-BUJLojNo4ik/UNIqWTKGwHI/AAAAAAAAAAY/DA4YD6EGJh4/s1600/world_erd.png

PR #96, opened 2018-05-09
initial commit

thwjason:
Not entirely sure if I understand what the homework is all about. 

I have:- 

1) done the drawing using LucidChart, not sure if this shows up if you click on it. Saved this in a .txt file.
2) Done the queries for the databases, and saved it as part of the same .txt file.

I have not:-

A) Done the bonus assignment. I will try to do it at some point this week.

### DIFF
--- a/Week1/homework/homework.txt
+++ b/Week1/homework/homework.txt
@@ -1,0 +1,81 @@
+Here is the lucidChart diagram:
+https://www.lucidchart.com/documents/edit/00de9f5d-4578-420f-8fed-93e057398424/0
+
+Here are the attempts at the database queries.
+
+1. What are the names of all the cities in the Netherlands?
+
+	SELECT city.Name 
+	FROM city 
+	JOIN country 
+	ON city.CountryCode = country.Code 
+	WHERE country.Name = 'Netherlands';
+
+2. What is the population of Rotterdam ?
+
+	SELECT population 
+	FROM city 
+	WHERE name='Rotterdam';
+
+3. What's the name of all the countries on the continent ‘North America’ ?
+
+	SELECT name 
+	FROM country 
+	WHERE continent='North America';
+
+4. What's the top 10 countries by SurfaceArea ?
+
+	SELECT name 
+	FROM country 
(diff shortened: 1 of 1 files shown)


Week1/homework/homework.txt
```diff
@@ -0,0 +1,81 @@
+Here is the lucidChart diagram:
+https://www.lucidchart.com/documents/edit/00de9f5d-4578-420f-8fed-93e057398424/0
```
njaradat: I tried to use the link, but I don't have access to it. Could you take a screenshot of it instead and push the image to git? Thanks!

Week1/homework/homework.txt
```diff
@@ -0,0 +1,81 @@
...
+	SELECT name 
+	FROM country 
+	WHERE continent='North America';
+
+4. What's the top 10 countries by SurfaceArea ?
+
+	SELECT name 
+	FROM country 
+	ORDER BY SurfaceArea DESC
+	[alternatively, "GROUP BY SurfaceArea DESC"]
```
njaradat: Interesting point! It's better to stick to `ORDER BY` like you did though, because the use of `group by` in that way is quite unconventional, so it might be hard for other developers to understand what you are doing and maintain your code. The performance of it is also worse. 
Fun fact we didn't talk about in class: you can add comments to SQL by putting `--` before the comment

Week1/homework/homework.txt
```diff
@@ -0,0 +1,81 @@
...
+ 	countrylanguage
+	
+	JOIN
+	city ON city.Countrycode = countrylanguage.Countrycode
+	
+	JOIN
+	country ON city.Countrycode = country.Code
+	
+	WHERE
+	city.Name = 'Barcelona';
```
njaradat: Queries look great :)

Week1/homework/homework.txt
thwjason: Let's see if this works - if not, I will use an image.

https://www.lucidchart.com/invitations/accept/9ed51ccc-14fb-470c-ac92-2d5cd9ec2440

PR #97, opened 2018-05-10
Zaher HW Week1 DB

zaherqasqas:


njaradat:
This comment is for the ER diagram:
You also need another entity for tags so that you can cover this point in the homework: "Add tags to the TODO items, like homework, groceries etc..".

Week1/homework/world-queries.txt
```diff
@@ -1,40 +1,362 @@
...
+USE `mydb` ;
+
+-- -----------------------------------------------------
+-- Table `mydb`.`Users`
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS `mydb`.`Users` (
+  `Id_User` INT NOT NULL AUTO_INCREMENT,
+  `FirstName_User` VARCHAR(45) NULL,
+  `LastName_User` VARCHAR(45) NULL,
+  `Email_User` VARCHAR(45) NULL,
```
njaradat: There is no need to add the table name to all the column names. The more common convention is not to use the table name.

Week1/homework/world-queries.txt
```diff
@@ -1,40 +1,362 @@
...
+| Istanbul              |    8787958 |
+| Ciudad de Mï¿½xico    |    8591309 |
+| Moscow                |    8389200 |
+| New York              |    8008278 |
++-----------------------+------------+
+10 rows in set (0.00 sec)
+
+---------------------------------------------------------------------------------
+6- Select the top 3 countries by population that their names start with the letter ‘P’?
+ SELECT country.name, country.population FROM country WHERE country.name LIKE 'p%' LIMIT 3;
```
njaradat: To get the top 3 countries by population you need to use the `order by` clause too:

`SELECT country.name, country.population FROM country WHERE country.name LIKE 'p%' ORDER BY population DESC LIMIT 3;`

Week1/homework/world-queries.txt
```diff
@@ -1,40 +1,362 @@
...
+	countrylanguage.language,
+	countrylanguage.IsOfficial, 
+	country.region 
+  FROM 
+	countrylanguage, 
+	country 
+  WHERE  
+	country.region = 'South America'
+  AND
+	countrylanguage.countrycode = country.code
```
njaradat: It's nice to see that you found a different way of doing joins. It is correct, but these days using an INNER JOIN and ON clause is more common, because it is considered more readable. If you have many conditions in the WHERE clause, the syntax you are using can become confusing.
This is an alternative answer to the question:

`SELECT distinct language 
FROM country c
JOIN countryLanguage cc ON c.code = cc.countryCode
WHERE region = 'South America';`

Week1/homework/world-queries.txt
```diff
@@ -1,40 +1,362 @@
...
+| Guaranï¿½    | T          | South America |
+| Ketï¿½ua     | T          | South America |
+| Portuguese   | T          | South America |
+| Spanish      | T          | South America |
++--------------+------------+---------------+
+6 rows in set (0.00 sec)
+
+---------------------------------------------------------------------------------
+9- What are the languages spoken on all cities named ‘Barcelona’ (you may need to join 3 tables =) )
+SELECT DISTINCT countrylanguage.language, city.name FROM countrylanguage, city WHERE CITY.NAME = 'Barcelona'AND countrylanguage.IsOfficial = 'T'ORDER BY countrylanguage.language;
```
njaradat: This is a good example of my comment on line 234 about how this form of joining can be confusing, here you forgot to add this condition ` AND countrylanguage.countryCode = city.countryCode `. I would recommend to use joins instead.

Also, you don't need `IsOfficial = 'T'` here because we want to see all the languages.

Week1/homework/world-queries.txt
```diff
@@ -1,40 +1,362 @@
...
+    ON DELETE NO ACTION
+    ON UPDATE NO ACTION)
+ENGINE = InnoDB;
+
+
+SET SQL_MODE=@OLD_SQL_MODE;
+SET FOREIGN_KEY_CHECKS=@OLD_FOREIGN_KEY_CHECKS;
+SET UNIQUE_CHECKS=@OLD_UNIQUE_CHECKS;
+---------------------------------------------------------------------------
+
```
njaradat: Great job with the queries!

PR #98, opened 2018-05-11
Murat database week1 homework

MURATKAYMAZ56:
murat database week 1 homework


njaradat:
> Design a database for an app that keeps track of TODO lists of its **users**

The ER diagram looks good. You need one more entity to represent the users of the app though. You will also need to add the relationship between the user and the Todo entity.

Week1/homework/world-queries.txt
```diff
@@ -37,4 +37,180 @@ SELECT city.Name FROM city JOIN country ON city.CountryCode = country.Code WHERE
...
+| Karachi           |
+| Istanbul          |
+| Ciudad de México  |
+| Moscow            |
+| New York          |
++-------------------+
+6. Select the top 3 countries by population that their names start with the letter ‘P’
+mysql> SELECT name
+    -> FROM country
+    -> WHERE name like "p%" ORDER BY population LIMIT 3;
```
njaradat: You want the 3 biggest countries here, so you should do `ORDER BY population DESC`. It's a good idea to do `SELECT name, population FROM ...` in this query so you can double check that the population really is ordered.

Week1/homework/world-queries.txt
```diff
@@ -37,4 +37,180 @@ SELECT city.Name FROM city JOIN country ON city.CountryCode = country.Code WHERE
...
+mysql> SELECT sum(population)
+    -> FROM country
+    -> WHERE continent="ASIA";
++-----------------+
+| sum(population) |
++-----------------+
+|      3705025700 |
++-----------------+
+8. Which languages are spoken in the Region of ‘South America’
+mysql> SELECT language
```
njaradat: You can use `SELECT DISTINCT language` here so that you don't see the same language more than once in the results (for example Spanish and Italian are repeated more than once in the results).

Week1/homework/world-queries.txt
```diff
@@ -37,4 +37,180 @@ SELECT city.Name FROM city JOIN country ON city.CountryCode = country.Code WHERE
...
+| Spanish          |
+| Hindi            |
+| Sranantonga      |
+| Spanish          |
+| Goajiro          |
+| Spanish          |
+| Warrau           |
++------------------+
+9. What are the languages spoken on all cities named ‘Barcelona’ (you may need to join 3 tables =) )
+mysql> SELECT language
```
njaradat: Here is another place where you could use the `distinct` keyword so that you don't see Spanish repeated twice in the results.

Week1/homework/world-queries.txt
```diff
@@ -37,4 +37,180 @@ SELECT city.Name FROM city JOIN country ON city.CountryCode = country.Code WHERE
 | Heerlen           |
 | Alkmaar           |
 +-------------------+
-28 rows in set (0.00 sec)
\ No newline at end of file
+28 rows in set (0.00 sec)
```
njaradat: Good job, the queries look great!

PR #99, opened 2018-05-12
Fadi week1 homework

Fadi-Rasheed:


Week1/homework/world-queries.txt
```diff
@@ -1,4 +1,4 @@
-Answers for the world queries go here:
+﻿Answers for the world queries go here:
```
njaradat: This comment is for the ER diagram because I can't comment on it directly:

- There's no need to split the user data into two entities here. You can remove the UserInfo entity and just add all the columns to Users. 
- You can remove the NumOfUnfinishedTodo column, because you can always figure out this number by counting the records in the TODO table, so there is no need to store this number.
- What would you use as the primary key in the TODO table? I would recommend adding an id to this entity because there are no unique columns.
- You need another entity for the tags so that you can cover this point in the homework:
"Add tags to the TODO items, like homework, groceries etc..".

PR #101, opened 2018-05-18
Week2

Rushkurdi:


Week2/todo_app/program.js
```diff
@@ -21,39 +21,90 @@ class TodoModel {
         });
     }
 
-    create(description, callback) {
+    create(userId, description, callback) {
         // Write code and query to create a new TODO item
+        const createTodoItem = `insert into todo_items (text,user_id) values(${mysql.escape(description)}, ${mysql.escape(userId)});`;
```
zhaobn: You can do this. To make your life even easier,`mysql.js` also provides a neater way of escaping with `?`s. You can find detailed explanations of it in the **Escaping query values** section in `mysql.js` official documentation.

PR #102, opened 2018-05-18
Homework Week 2

WaelAlnadeem:


Week2/todo_app/program.js
```diff
@@ -21,39 +22,96 @@ class TodoModel {
         });
     }
 
-    create(description, callback) {
-        // Write code and query to create a new TODO item
+    create(user_id, description, callback) {
+        const createTodoItem = `INSERT INTO todo_items(text, is_completed, user_id)
+        VALUES(${dbConnection.escape(description)}, 0, ${dbConnection.escape(user_id)})`;
```
zhaobn: Good to use escaping. There's a even neater way of escaping with `?`s - you can find it under the **Escaping query values** section in `mysql.js` official documentation :)

Week2/todo_app/program.js
```diff
@@ -62,11 +120,64 @@ dbConnection.connect(function(err) {
...
+        if (err) {
+            console.log("error loading TODO items:", err);
+        }
+
+        console.log("existing todo items:", todoItems);
+    });
+
+
+    // // create
+    todoModel.create(2, "create anything", function (err, todoItems) {
```
zhaobn: We'd like our app to be interactive - get input from users and function really as an app. Can you think about how to do that?

PR #103, opened 2018-05-19
Murat database week2 hw

MURATKAYMAZ56:
database week 2 added

### DIFF
--- a/Week2/todo_app/program.js
+++ b/Week2/todo_app/program.js
@@ -11,8 +11,8 @@ class TodoModel {
     // Loads all the TODOs in the database
     load(callback) {
         const selectTodoItems = "SELECT * FROM todo_items";
-        this.dbConnection.query(selectTodoItems, function(err, results, fields) {
-            if(err) {
+        this.dbConnection.query(selectTodoItems, function (err, results, fields) {
+            if (err) {
                 callback(err);
                 return;
             }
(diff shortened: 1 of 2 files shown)


Week2/todo_app/program.js
```diff
@@ -23,37 +23,98 @@ class TodoModel {
 
     create(description, callback) {
         // Write code and query to create a new TODO item
+
+        this.dbConnection.query('INSERT INTO todo_items (text, user_id) VALUES(?,?)', [description, 4], function (err, results, fields) {
```
zhaobn: You can do this - but it's better to make your query string a variable instead of putting it directly there. Just a suggestion :) I see that you do it here because our queries are short, when you encounter huge queries in the future, better not do this.

Week2/todo_app/program.js
```diff
@@ -62,11 +123,20 @@ dbConnection.connect(function(err) {
...
-            console.log("error loading TODO items:", err);
+    // todoModel.load(function (err, todoItems) {
+    //     if (err) {
+    //         console.log("error loading TODO items:", err);
+    //     }
+
+    //     console.log("existing todo items:", todoItems);
+    // });
+
+    todoModel.delete(46, function (err, todoItems) {
```
zhaobn: We'd like to make this app interactive - get input from the terminal and test all the functionalities you just made. Can you make this happen?

PR #105, opened 2018-05-20
Aya week2 hm

Aya-Alabrash:


Week2/todo_app/program.js
```diff
@@ -4,69 +4,195 @@
...
-                callback(err);
-                return;
-            }
-
-            callback(null, results);
-        });
-    }
+  // Loads all the TODOs in the database
+  load(callback) {
+    const selectTodoItems = "SELECT * FROM todo_items";
```
zhaobn: why did you change the indentation?

Week2/todo_app/program.js
```diff
@@ -4,69 +4,195 @@
...
+    });
+  }
 
-    update(id, description, callback) {
-        // Write code and query to update and existing TODO item
-    }
+  create(description, callback) {
+    // Write code and query to create a new TODO item
+    const createTodoItem = `INSERT INTO todo_items(text, is_completed, user_id) 
+    VALUES (${dbConnection.escape(description)}, 0, ${dbConnection.escape(arg2)} )`;
```
zhaobn: Better to use `id` from your input as `user_id`. This works now because you defined a global variable `arg2` on line 107, but this is not a good practice if you want to change they way you name your variables one day in the future.

PR #111, opened 2018-05-30
Sarbast's  week3 HW

SarKurd:


week3/server.sql
```diff
@@ -0,0 +1,40 @@
+CREATE TABLE `customer` (
```
njaradat: Comments on ER diagram:

- There should be a one to many relationship between product and cart. Every record of cart contains only one product. A better name for the cart entity would be "customer_product" or "item_in_cart", the name "cart
 makes it sound like there is more than one product in it, but in reality this table will only hold one product in every record.
- The relationship between cart and order needs improvement. You have a many to many relationship, but no junction table. You should have a junction table instead of the foreign_key cart_id. In the current setup with this foreign key, every order can only contain one cart (but if a customer is buying 20 items, the order should contain 20 carts). Again, I think the confusion comes from the naming of the entity cart.

week3/server.sql
```diff
@@ -0,0 +1,40 @@
...
+  `ID` INT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `f_name` varchar(35) NOT NULL,
```
njaradat: I would encourage you to use complete words when naming columns. f_name could mean full_name or first_name. For people who are using the database after you, it is always easier to understand when there are no abbreviations.

week3/server.sql
```diff
@@ -0,0 +1,40 @@
...
+  `l_name` varchar(35) NOT NULL,
+  `email` varchar(255) NOT NULL,
+  `phone_number` varchar(45) NOT NULL,
+  `Address` varchar(255) NOT NULL
```
njaradat: Using all lowercase letters is better because some database management systems are case sensitive. If you ever need to move your database from MySQL to another database, this could introduce errors.

week3/server.sql
```diff
@@ -0,0 +1,40 @@
...
+  `customer_id` INT NOT NULL,
+  FOREIGN KEY (`product_id`) REFERENCES `product` (`ID`),
+  FOREIGN KEY (`customer_id`) REFERENCES `customer` (`ID`)
+);
+
+CREATE TABLE `order` (
+  `ID` INT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `Date` datetime NOT NULL,
+  `cart_id` int NOT NULL,
+  FOREIGN KEY (`cart_id`) REFERENCES `cart` (`ID`)
```
njaradat: I mentioned this in the ER diagram comments too. This constraint makes this into a one to many relationship, where 1 cart can be in many orders, but 1 order only contains one cart. And if 1 order only contains one cart, that means it only contains one product, which is incorrect in real life.

week3/server.sql
```diff
@@ -0,0 +1,40 @@
...
+);
+
+CREATE TABLE `shipping` (
+  `ID` INT NOT NULL AUTO_INCREMENT PRIMARY KEY,
+  `order_id` INT NOT NULL,
+  `date` datetime NOT NULL,
+  FOREIGN KEY (`order_id`) REFERENCES `order` (`ID`)
+);
+
+
```
njaradat: You translated all of the ER diagram into SQL accurately, good job!

SarKurd:
Will apply your feedback soon and update the PR

PR #112, opened 2018-05-30
HomeWork

WaelAlnadeem:


gijscor:
Dear Wael, can you please add the week of the homework to the name as asked?
Thanks!

WaelAlnadeem:
Oh , sorry I didn’t notice that I didn’t put the week , I will change it as soon as I  reach HYF

Week3/Hotel.sql
```diff
@@ -0,0 +1,96 @@
+-- MySQL Workbench Forward Engineering
```
njaradat: Try to write this SQL from scratch instead of using workbench to generate it,

Week3/Hotel.sql
```diff
@@ -0,0 +1,96 @@
...
+
```
njaradat: Comments on ER diagram:

- I need more context to be able to understand the ER diagram. What is the business that you are building this database for? A hotel?

Week3/Hotel.sql
```diff
@@ -0,0 +1,96 @@
...
+-- Schema mydb
+-- -----------------------------------------------------
+CREATE SCHEMA IF NOT EXISTS `mydb` DEFAULT CHARACTER SET utf8 ;
+USE `mydb` ;
+
+-- -----------------------------------------------------
+-- Table `mydb`.`Employees`
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS `mydb`.`Employees` (
+  `idEmployees` INT NOT NULL,
```
njaradat: You can use the name "id" here instead of "idEmployees", it is already implied that the id is for an employee since it is in the employee table.

Week3/Hotel.sql
```diff
@@ -0,0 +1,96 @@
...
+  `username` VARCHAR(45) NOT NULL,
+  `Password` VARCHAR(45) NULL,
```
njaradat: Try to keep the same naming conventions throughout your tables. It's better to have all lowercase letters to avoid confusion.

Week3/Hotel.sql
```diff
@@ -0,0 +1,96 @@
...
+-- -----------------------------------------------------
+-- Table `mydb`.`customer`
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS `mydb`.`customer` (
+  `idcustomer` INT NOT NULL AUTO_INCREMENT,
+  `first name` VARCHAR(45) NOT NULL,
+  `last name` VARCHAR(45) NOT NULL,
+  `birthday` INT NOT NULL,
+  `customer-num` INT NOT NULL,
+  `Employees_idEmployees` INT NOT NULL,
```
njaradat: This column is hard to understand, which employee should be stored here? 
- Is it the employee that entered this customer's data into the system? A better column name would be: created_by
- Is it the employee who cleans their room? A better column name would be housekeeper_id
- Is it the employee that is responsible for them throughout their stay? 

The name would also help me to understand what the correct relationship would be in the ER diagram.

Week3/Hotel.sql
```diff
@@ -0,0 +1,96 @@
...
+  UNIQUE INDEX `username_UNIQUE` (`username` ASC))
+ENGINE = InnoDB;
+
+
+-- -----------------------------------------------------
+-- Table `mydb`.`customer`
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS `mydb`.`customer` (
+  `idcustomer` INT NOT NULL AUTO_INCREMENT,
+  `first name` VARCHAR(45) NOT NULL,
```
njaradat: It's better not to use spaces in column names, if you try to query this you will always have to put back quotes around the words 'first name' for the select statement to work. It will assume the column name is done after you write the word `first`. Try it on your database.

Week3/Hotel.sql
```diff
@@ -0,0 +1,96 @@
...
+  `last name` VARCHAR(45) NOT NULL,
+  `birthday` INT NOT NULL,
+  `customer-num` INT NOT NULL,
```
njaradat: - Is this similar to an id? or is this their telephone number? In that case, a better column name would be `telephone`. 
- Also try to avoid using dashes in column names, they can cause code to break at times. Use underscores instead.

Week3/Hotel.sql
```diff
@@ -0,0 +1,96 @@
...
+CREATE TABLE IF NOT EXISTS `mydb`.`rooms` (
+  `idrooms` INT NOT NULL AUTO_INCREMENT,
+  `room-number` INT NOT NULL,
+  `price` INT NULL,
+  `date_of_rent` DATE NOT NULL,
+  `end_day` DATE NOT NULL,
+  `empty_or_not` TINYINT NULL,
+  `Employees_idEmployees` INT NOT NULL,
+  `room_type` VARCHAR(45) NULL,
+  `roomscol` VARCHAR(45) NOT NULL,
```
njaradat: What is this column used for?

Week3/Hotel.sql
```diff
@@ -0,0 +1,96 @@
...
+ENGINE = InnoDB;
+
+
+-- -----------------------------------------------------
+-- Table `mydb`.`rooms`
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS `mydb`.`rooms` (
+  `idrooms` INT NOT NULL AUTO_INCREMENT,
+  `room-number` INT NOT NULL,
+  `price` INT NULL,
```
njaradat: The datatype for this should be decimal not int. The price could be 20.5 euros.

Week3/Hotel.sql
```diff
@@ -0,0 +1,96 @@
...
+-- Table `mydb`.`customer`
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS `mydb`.`customer` (
+  `idcustomer` INT NOT NULL AUTO_INCREMENT,
+  `first name` VARCHAR(45) NOT NULL,
+  `last name` VARCHAR(45) NOT NULL,
+  `birthday` INT NOT NULL,
+  `customer-num` INT NOT NULL,
+  `Employees_idEmployees` INT NOT NULL,
+  PRIMARY KEY (`idcustomer`, `Employees_idEmployees`),
```
njaradat: The primary key should only be idcustomer here. Employees_idEmployees should be removed.

Week3/Hotel.sql
```diff
@@ -0,0 +1,96 @@
...
+
+
+-- -----------------------------------------------------
+-- Table `mydb`.`customer`
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS `mydb`.`customer` (
+  `idcustomer` INT NOT NULL AUTO_INCREMENT,
+  `first name` VARCHAR(45) NOT NULL,
+  `last name` VARCHAR(45) NOT NULL,
+  `birthday` INT NOT NULL,
```
njaradat: The datatype should be date not int

PR #113, opened 2018-06-01
Emad week2

EmadMRI:


Week2/todo_app/program.js
```diff
@@ -21,39 +21,97 @@ class TodoModel {
         });
     }
 
-    create(description, callback) {
+    create(description, userId, callback) {
         // Write code and query to create a new TODO item
+        const createTodoItem = `INSERT INTO todo_items(text, is_completed, user_id) 
+                                VALUES (${dbConnection.escape(description)}, "0", ${dbConnection.escape(userId)})`;
```
zhaobn: Good to use escape. There's also prepared statement provided by mysql.js to make your node more neat. Do check it out ;)

zhaobn: Did you try whether `"0"` works? Isn't the `is_completed` column has boolean/int data type?

Week2/todo_app/program.js
```diff
@@ -62,11 +120,87 @@ dbConnection.connect(function(err) {
...
+                    }
+                    console.log('Todo item has been completed successfully')
+                });
+            }
+
+        default:
+            console.log("404 error: commend Not found!");
+            break;
+    }
+
```
zhaobn: Always remember to close your dbconnection after querying it :)

PR #123, opened 2018-07-19
Meazer week1 hw

Meazer:


Meazer:
Hi Frank,
Thanks for your feedback.

PR #126, opened 2018-07-19
Week1 homework

zmagzoub:


zmagzoub:
Hello Frank,

Thank you for your feedback. I've updated my homework and I think now query 6 is correct.

Cheers,
Zaed

PR #127, opened 2018-07-19
database homework week 1

amirno1:


amirno1:
Hi Frank, 

I was not present on last sunday lecture and only watched the recoroded version which some parts are being explaind on the board and i could not really understand what was going on. When i tried to do homeworks i had no idea about the part with TODO database design but i will try my best to learn it by myself and compelete the homework.  That's why i asked on the slack to try not to use the board if it is possible cause i can not attend the next lecture as well. 

Thanks,
Amir

PR #128, opened 2018-07-19
Added homework week1

neveenatik:
Thanks in advance for your feedback. I really appreciate that!


neveenatik:
Hi Frank,
Thanks for your feedback answer was indeed incorrect now it is corrected.

PR #132, opened 2018-07-26
Added homework week2

neveenatik:


neveenatik:
Hi Frank,

That's right, I changed userId to make it a requested parameter. So it needs to be provided.

Thanks for your review.

PR #133, opened 2018-07-27
meazer week2 homework

Meazer:


Meazer:
Hi Frank, 
Thanks for your review.

Here the affectedRows will be equal to zero in cases like (delete, update or markCompleted with non existing item_id. and tagTodoItem or untagTodoItem with non existing ids) so I ask the user to check the given values !.
if there is a zero Todo items the program will return []. cause load function doesn't call _query function.

PR #144, opened 2018-08-10
Homework week1

ThreePalmTrees:


fversnel:
Hi Husam,
Queries look good, well done. About the TODO database design a few remarks:
1. What is the help table for, to what concept does it relate in the homework description?
2. What does the `state` column in your Todo table describe. Can you think of a better, less general, name for the column?

ThreePalmTrees:
@fversnel Thank you Frank, really appreciate the feedback,
1. My idea behind the help table is that if a user finds a task hard to finish by him/her self then he/she can ask for help, it doesn't relate to the homework description but I thought it would be a nice addition.
2. The `state` column describe the todo state, if it's done or not done.. Will `is_done` be a better name instead ?

fversnel:
1. Alright, that's fair enough but from the diagram alone I couldn't figure that out.
2. What do you think is a better name? The dictionary tells us that state means: "the particular condition that someone or something is in at a specific time." Does that accurately describe the information that you want to store? (To give you a hint: I think state is very general and can mean a lot of different *specific* things)

PR #148, opened 2018-09-23
homework finished

end0620:


Week1/homework/world-queries.sql
```diff
@@ -0,0 +1,45 @@
...
+LIMIT 10;
+
+#Select the top 3 countries by population that their names start with the letter ‘P’
+SELECT country.Name, country.Population FROM country 
+WHERE country.Name LIKE 'P%' 
+ORDER BY country.Population DESC 
+LIMIT 3;
+
+#What is the population in Asia?
+SELECT country.Continent, SUM(country.Population) AS AsiaPopulation FROM country
```
pcelentano: nice use of aliases 👍

Week1/homework/world-queries.sql
```diff
@@ -0,0 +1,45 @@
...
+WHERE country.Continent = 'Asia' 
+GROUP BY country.Continent;
+
+#Which languages are spoken in the Region of ‘South America’
+SELECT country.Name, countrylanguage.Language FROM country 
+JOIN countrylanguage ON country.Code = countrylanguage.CountryCode 
+WHERE country.Region = 'South America';
+
+#What are the languages spoken on all cities named ‘Barcelona’
+SELECT city.Name, countrylanguage.Language FROM city
```
pcelentano: 👏

Week1/homework/todo sql.sql
```diff
@@ -0,0 +1,42 @@
+CREATE DATABASE IF NOT EXISTS  base_todoinfo;
+
+USE base_todoinfo;
+
+CREATE TABLE IF NOT EXISTS todos(
+	id_todo INT NOT NULL,
```
pcelentano: (this was not part of the homework... but while we are at it) 
convention is we only call it id, any time it will get referenced it will be prefixed by `todos`. `id_todo` looks like a forign key.. which we normally use `todo_id`.  If we retrieve a single row its a todo, so if we want the id its just `todo.id` or if we use the id to join table we write the whole `on todos.id = other_entity.todo_id`

pcelentano: oh and while we are at it:
`id int NOT NULL AUTO_INCREMENT ` -> to get the auto increment id (well see how this helps when inserting data next class)

Week1/homework/todo sql.sql
```diff
@@ -0,0 +1,42 @@
...
+    dead_line DATE,
+    done BOOLEAN,
+    PRIMARY KEY(id_todo)
+);
+CREATE TABLE IF NOT EXISTS users(
+	id_user INT NOT NULL,
+    name VARCHAR(25),
+    PRIMARY KEY(id_user),
+    todos_id_todo INT NOT NULL,
+    CONSTRAINT fkuser_todos
```
pcelentano: this constraint here implies the relation between users and todos is one user has one todo and one todo has many users 🤔  If you did this one from a sql generatior make sure the `many` sides of the relationship chicken wings is pointing to the correct table

end0620: ohh yes sorry, the idea from the start was to make one user to have many todos, u put the fk in the wrong place :)

end0620:
@pcelentano  what do you mean to make my todo app support multiple users?

unmeshvrije:
Please re-open the PR after you make changes

PR #161, opened 2018-10-03
Shadi week2

shadoo77:


Week2/todo_app/program.js
```diff
@@ -23,36 +37,141 @@ class TodoModel {
...
+        allItemsId
+            .then(data => {
+                const todoIdIsFound = data.includes(todoItemId);
+                if(todoIdIsFound) {
+                    const checkTagId = this.availableIds("SELECT * FROM tags");
+                    checkTagId
+                        .then(tagData => {
+                            const tagIdIsFound = tagData.includes(tagId);
+                            if(tagIdIsFound) {
+                                const addTodoItemTag = "INSERT INTO todo_item_tag VALUES(?, ?)";
```
pcelentano: the previous checks are a little redundant and NOT needed on sql, remember this is not a procedural language, its a declarative one. You query all the data and checking in memory if an id exists, the database can do all of this operations way faster and without using the memory. 

you could have checked this in sql with something like 
```select count(1)
from table
where key = value;```

but its not even needed because of the foreign key constraints, if the foreign keys don't exist the record won't be inserted

File: Week2/todo_app/program.js [hunk 2]
```diff
@@ -23,36 +37,141 @@ class TodoModel {
...
+        allItemsId
+            .then(data => {
+                const todoIdIsFound = data.includes(todoItemId);
+                if(todoIdIsFound) {
+                    const checkTagId = this.availableIds("SELECT * FROM tags");
+                    checkTagId
+                        .then(tagData => {
+                            const tagIdIsFound = tagData.includes(tagId);
+                            if(tagIdIsFound) {
+                                const addTodoItemTag = "DELETE FROM todo_item_tag WHERE todo_item_id = ? AND tag_id = ?";
```
pcelentano: once again all previous checks are not needed

Week2/todo_app/program.js
```diff
@@ -23,36 +37,141 @@ class TodoModel {
...
     }
 
     markCompleted(todoItemId, callback) {
         // Write code to mark a TODO item as completed
+        const updatePromise = this.availableIds(this.sqlSelectQuery);
+        updatePromise
+            .then(data => {
+                const temp = data.includes(todoItemId);
+                if(temp) {
+                    const updateItem = `UPDATE todo_items SET is_completed = ? WHERE id = ?`;
```
pcelentano: same 😜

Week2/todo_app/program.js
```diff
@@ -62,11 +181,46 @@ dbConnection.connect(function(err) {
...
+
+    todoModel.tagTodoItem(44, 2, (err) => {
+        err ? console.log("Error by adding new tag", err) : console.log("New todo-item tag added!");
+    });
+
+    todoModel.untagTodoItem(44, 2, (err) => {
+        err ? console.log("Error by deleting the tag", err) : console.log("tag deleted !");
+    });
+    
+    todoModel.markCompleted(43, (err) => {
```
pcelentano: loved the tests, you can query the db on the mysql terminal to see that the db ends up populated like you intended 👍

PR #162, opened 2018-10-03
finished homework week2

genegr88:
Wanted to do this assignment with express and endpoints but a bit time-consuming....so a bit simplified version of my homework

### DIFF
--- a/Week2/todo_app/program.js
+++ b/Week2/todo_app/program.js
@@ -4,69 +4,241 @@
 const mysql = require('mysql');
 
 class TodoModel {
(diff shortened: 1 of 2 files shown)


Week2/todo_app/program.js
```diff
@@ -4,69 +4,241 @@
...
-    });
-});
\ No newline at end of file
+      });
+      break;
+    }
+    case 'help':
+    default:
+      {
+        console.info(
```
pcelentano: nice testing and documentation ! 👍 🔥

PR #185, opened 2018-12-06
week1

joudi6:


Week1/h.w_select.js
```diff
@@ -0,0 +1,65 @@
...
+  if (error) throw error.message;
+  console.log('countries ordered descending by their areas:\n ', results, '\n');
+});
+
+connection.query('SELECT city_name  FROM cities where country_name like"Netherlands"', function (error, results, fields) {
+  if (error) throw error.message;
+  console.log('cities in the netherlands:\n ', results, '\n');
+});
+
+connection.query('SELECT city_name, population  FROM cities where city_name like"Rotterdam"', function (error, results, fields) {
```
unmeshvrije: Please notice the difference between like and = before using like. In this case, it does not matter, however there are cases where "like" and "=" give different results

PR #186, opened 2018-12-06
week1

karamad10:


Week1/home-work/queries.js
```diff
@@ -0,0 +1,95 @@
...
+   connection.query(result, function (error, result) {
+      if (error) throw error;
+      else if (result == 0) console.log('No countries with population between 500,000 and 1 million')
+      else console.log(result);
+   })
+};
+
+
+function euroCountries() {
+   var result = "SELECT countries.country_name, countries.continent FROM countries WHERE countries.continent LIKE '%europe%'";
```
unmeshvrije: %europe% is not entirely correct. '=' can be used to check equality. If there is a continent 'Teurope' by mistake, it will be matched.

Week1/home-work/queries.js
```diff
@@ -0,0 +1,95 @@
...
+function descendingOrder() {
+   var result = "SELECT * FROM countries ORDER BY countries.country_size DESC";
+   connection.query(result, function (error, result) {
+      if (error) throw error;
+      console.log("countries in descending order\n", result)
+   })
+};
+
+function citiesInNetherlands() {
+   var result = "SELECT * FROM cities WHERE cities.country LIKE '%Netherlands%'";
```
unmeshvrije: LIKE %Netherlands% is not the same as = 'Netherlands'.

Week1/home-work/queries.js
```diff
@@ -0,0 +1,95 @@
...
+function mostPopulatedCities() {
+   var result = "SELECT * FROM cities GROUP BY city_population ORDER BY city_population DESC";
+   connection.query(result, function (error, result) {
+      if (error) throw error;
+      console.log("most populated cities", result)
+   })
+};
+
+function worldPopulation() {
+   var result = "SELECT SUM(city_population) AS cities_population, SUM(country_population) AS world_population FROM cities INNER JOIN countries ON city_population + country_population";
```
unmeshvrije: I don't understand this one. Why is sum of all countries's population not enough for the world's population ?
If your database contains cities but not countries that contain those cities, then the database is incomplete.

PR #188, opened 2018-12-07
Here is the homework for week1

kawaraa:


Week1/create-database.js
```diff
@@ -0,0 +1,48 @@
...
+  password: "hyfpassword",
+  database: "world"
+});
+
+connection.connect();
+
+const queries = [
+  "CREATE DATABASE world",
+  "CREATE TABLE country (name VARCHAR(20) PRIMARY KEY, population INT, surface_area INT, continent VARCHAR(20));",
+  "CREATE TABLE city (name VARCHAR(20), population INT, country VARCHAR(20), FOREIGN KEY(country) REFERENCES country(name));"
```
unmeshvrije: good use of foreign key. can country name indeed be a primary key ?

Week1/create-database.js
```diff
@@ -0,0 +1,48 @@
...
+  host: "localhost",
+  user: "hyfuser",
+  password: "hyfpassword",
+  database: "world"
+});
+
+connection.connect();
+
+const queries = [
+  "CREATE DATABASE world",
```
unmeshvrije: what if the database already exists ?

kawaraa:
Hi @unmeshvrije Thank you for comments, sorry for late reply I did not receive an Email or any navigation about it!

PR #189, opened 2018-12-07
homework week1 DB

RimSebai:


myhw-week1/created-data.js
```diff
@@ -0,0 +1,181 @@
...
+connection.query(sql, function (error, results) {
+  if (error) {
+    throw error;
+  }
+  console.log(results);
+});
+
+connection.query("use world")
+
+let countries_query = "create table if not exists countries(country_name varchar(20) unique, country_population bigint, continent varchar(20), surface_area int,primary key(country_name))"
```
unmeshvrije: can country_name be a primary key ?

myhw-week1/db.js
```diff
@@ -0,0 +1,103 @@
...
+let all_countries = "select * from countries order by surface_area desc"
+connection.query(all_countries, function (error, results) {
+  if (error) {
+    throw error;
+  }
+  console.log(results);
+});
+
+// Answer number (6)
+let netherlands_cities = "select countries.country_name, cities.city_name from cities inner join countries on cities.country = countries.country_name where countries.country_name = 'netherlands'"
```
unmeshvrije: please explain the use of inner join.

myhw-week1/created-data.js
RimSebai: yes why not it is uniquely identifies  and contains UNIQUE values, right?

myhw-week1/db.js
RimSebai: I wanted to join two tables 
first i did specify which columns i want to display by using "select". 
second I used "from" to specify the tables i wanted to join. 
third I did specify which columns I wanted to much by using "on"  and the condition by using "where"

RimSebai: ooh I had to use join to get only the matched data.

myhw-week1/created-data.js
unmeshvrije: I will accept it for this homework.

myhw-week1/db.js
unmeshvrije: Good explanation! I like that you know what you are doing exactly.

PR #190, opened 2018-12-08
Week1 from Rabia

Damla78:


Week1/create-table.js
```diff
@@ -1,19 +1,34 @@
...
+  host: 'localhost',
+  user: 'hyfuser',
+  password: 'hyfpassword',
+  database: 'class17'
 });
 
 connection.connect();
 
-var create_query = "create table teachers (teacher_number int, teacher_name varchar(50), date_of_birth date, subject text, gender enum('m', 'f'))"
+var create_query_world = "CREATE TABLE IF NOT Exits` world (world_number int, world_name varchar(50), PRIMARY KEY (world_number) ) ";
```
unmeshvrije: The assignment asked to create "database" world not the "table"

Week1/insert-values.js
```diff
@@ -1,26 +1,72 @@
...
 });
 
 connection.connect();
 var insert_queries = [
-                    "insert into students values (103, 'Ahmad', '1986-01-11', 8.3, 'm')",
-                    "insert into students values (104, 'Rim', '1979-11-01', 8.7, 'f')",
-                    "insert into students values (105, 'Rabia', '1978-05-30', 8.1, 'f')",
-                    "insert into students values (106, 'Karam', '1994-02-02', 8.8, 'm')"
-                    ]
+  "insert into world values (1, 'Africa')",
```
unmeshvrije: "Treatment of continents as worlds" is an interesting idea but not what was asked for.
"Continent" should have been the field for the country table.

Week1/sql.js
```diff
@@ -0,0 +1,68 @@
...
+queryList[0] = "select * from countries where population>8000000";
+
+questions[1] = "What are the names of countries that have “land” in their names ? ";
+queryList[1] = "select country_name from countries where country_name LIKE '%land%'";
+
+questions[2] = "What are the names of the cities with population in between 500, 000 and 1 million ? ";
+queryList[2] = "select city_name from cities where(population > 500000 and population < 1000000)";
+
+questions[3] = "What's the name of all the countries on the continent ‘Europe’ ?";
+queryList[3] = "select country_name from countries where world_number = 6";
```
unmeshvrije: The number 6 should be deduced by the query. It should not be entered by the programmer

Week1/sql.js
```diff
@@ -0,0 +1,68 @@
...
+
+questions[4] = "List all the countries in the descending order of their surface areas.";
+queryList[4] = "select city_name from cities ORDER BY population DESC";
+
+questions[5] = "What are the names of all the cities in the Netherlands ? ";
+queryList[5] = "select city_name from cities, countries where countries.country_name = 'Nederland' and countries.country_number = cities.country_number";
```
unmeshvrije: How can you make sure that the city must have a country number that exists in the country table ?

PR #192, opened 2018-12-14
Week2

kawaraa:


Week2/part1.js
```diff
@@ -0,0 +1,68 @@
...
+  password: "hyfpassword",
+  database: "new_world"
+});
+
+const execQuery = util.promisify(con.query.bind(con));
+
+async function queryPromise() {
+
+  function capitalOfCountry(x) {
+    return `SELECT country.name, city.name FROM city, country WHERE id = capital AND country.name = ${x};`;
```
lao: I really like your separation with functions. =)

But, for security reasons, you should avoid concatenating input strings directly into the query like you did. Specially if the input comes directly from the user. In your case they are all hardcoded so it is kind of fine, but be aware of this. =)

Try one of the following approaches:

```
connection.query('SELECT * FROM users WHERE id = ?', [userId], function (error, results, fields) {
  if (error) throw error;
  // ...
});
```

or 

```
var userId = 'some user provided value';
var sql    = 'SELECT * FROM users WHERE id = ' + connection.escape(userId);
connection.query(sql, function (error, results, fields) {
  if (error) throw error;
  // ...
});
```

Or take a look at this link. 
https://github.com/mysqljs/mysql#escaping-query-values

Week2/part2.js
```diff
@@ -0,0 +1,113 @@
...
+        console.log(result[el]);
+      }
+    });
+  }
+
+  con.connect(err => console.log(err ? err.stack : "connected..."));
+
+  try {
+    for (let i = 0; i < languages.length; i++) {
+      let result = await execQuery("INSERT INTO countryLanguage SET ?", languages[i]);
```
lao: In this approach, you are safe from SQL injection attacks, for example. =)

Week2/part1.js
```diff
@@ -0,0 +1,68 @@
...
+      for (let el in result) {
+        console.log(result[el]);
+      }
+    });
+  }
+
+  con.connect(err => console.log(err ? err.stack : "connected..."));
+
+  try {
+    const result1 = await execQuery(capitalOfCountry("'Netherlands'"));
```
lao: With this other way, you would not need to have single quotes within double quotes:
```
connection.query('SELECT * FROM users WHERE id = ?', [userId], function (error, results, fields) {
  if (error) throw error;
  // ...
});
```

Week2/part1.js
```diff
@@ -0,0 +1,68 @@
...
+  user: "hyfuser",
+  password: "hyfpassword",
+  database: "new_world"
+});
+
+const execQuery = util.promisify(con.query.bind(con));
+
+async function queryPromise() {
+
+  function capitalOfCountry(x) {
```
lao: Maybe a next step in this part1 would be to take the inputs from the user? =)

Week2/part2.js
```diff
@@ -0,0 +1,113 @@
...
+    console.log(err.message);
+  }
+
+  con.end(err => console.log(err ? err : "disconnected."));
+}
+
+
+main();
+
+// languages.forEach(async lang => {
```
lao: What is this for? =)

Week2/part2.js
```diff
@@ -0,0 +1,113 @@
...
+  con.connect(err => console.log(err ? err.stack : "connected..."));
+
+  try {
+    for (let i = 0; i < languages.length; i++) {
+      let result = await execQuery("INSERT INTO countryLanguage SET ?", languages[i]);
+      console.log(result);
+      // printAnswer(result, 1);
+    }
+  } catch (err) {
+    console.log(err.message);
```
lao: Suggestion: You could also improve your message handling by getting the SQLSTATE so you could handle that error in a different way than other errors.

https://github.com/mysqljs/mysql#error-handling

Week2/part3.js
```diff
@@ -0,0 +1 @@
+//
```
lao: What about the last part?

kawaraa:
Thanks for your great comments. and sorry for late reply!

PR #194, opened 2018-12-14
Week2

RimSebai:


myhw-week1/created-data.js
```diff
@@ -0,0 +1,181 @@
...
+  user: 'hyfuser',
+  password: 'hyfpassword',
+  database: 'userdb'
+});
+
+connection.connect();
+
+// creating database world
+let sql = "create database if not exists world"
+connection.query(sql, function (error, results) {
```
lao: Query method is an asynchronous call. You might use this example as a guide for taking this into account: https://github.com/HackYourFuture/databases/blob/master/Week2/async-create-insert.js =)

myhw-week2/database.js
```diff
@@ -0,0 +1,73 @@
...
+  } catch (error) {
+    console.error(error);
+  }
+
+  connection.end();
+}
+seedDatabase();
+
+
+// part 2
```
lao: This part 2 could also be done with nodejs =)

myhw-week2/database.js
```diff
@@ -0,0 +1,73 @@
...
+});
+
+
+const execQuery = util.promisify(connection.query.bind(connection));
+
+async function seedDatabase() {
+
+
+  //Part 1 - 1
+  let country_name = 'Syria';
```
zhaobn: The idea of this homework is to take input from the user, not hardcoded in the code :) Consider using `process.argv` for this purpose.

myhw-week2/database.js
```diff
@@ -0,0 +1,73 @@
...
+  let capitalOfCountry = 'select country.name, city.name from city inner join country on city.id = country.capital where country.name = ' + connection.escape(country_name);
```
zhaobn: Good practice to use `connection.escape` 👍

myhw-week2/database.js
```diff
@@ -0,0 +1,73 @@
...
+
+  //Part 1 - 2
+  let country_region = 'Western Europe';
+  let findingSpokenLanguages = 'select countryLanguage.language, country.region from country join countryLanguage on country.Code = countryLanguage.countryCode where country.region =' + connection.escape(country_region);
```
zhaobn: One way to write less code is to give your tables alias, for example `countryLanguage as l` and then you can join on `l.countryCode` ;)

PR #199, opened 2018-12-15
week2

joudi6:
sorry for late handing.
- part 1.4 select statement still unfinished
- I found a nice npm prompt to get user input but will need more time to chose a good way to apply it with many functions


lao:
Nice that you found a good prompt package from npm. I would suggest you to also take a look at your colleagues PRs and maybe try this package. Would be a good exercise. =)

PR #200, opened 2018-12-17
week 3 content and homework

unmeshvrije:


Week3/README.md
```diff
@@ -1,37 +1,96 @@
...
+Subject table (subject_ID, subject Name)
+```
+#### 3NF
+No transitive dependency (i.e. no field should depend on non-key attributes).
+
+#### Boyce Codd Normal Form (3.5 NF)
+for any dependency A → B, A should be a super key.
+
+#### 4NF
+No multi-value dependency.
```
remarcmij: These normalization rules as formulated here seem a bit much for our students to take in at this stage. An alternative might be to consider explaining what normalization is and why it is important in simpler terms here and link to some external resource for details?

Week3/README.md
```diff
@@ -1,37 +1,96 @@
...
+# Lesson 3: Database design, normal forms, SQL injection
 
-In the final week, additional theory will be covered to discuss more complex relational data. Students will learn about entity relationship modelling and how to convert these models to a database schema using normalisation and foreign-key constraints. Non-relational data will also be considered, as well as the benefits and drawbacks of relational and non-relational models.
-
-Objective: Students should be able to create an entity relationship diagram based on a qualatative description of data requirements, and translate that into a MySQL database schema. Students should also be able to compare and contrast relational (like MySQL) and NoSQL databases (considering their benefits and drawbacks).
+Objective: This class invites students to discuss Entity Relationship Diagram (ERD).
+Students should be able to explain their choices of entities, relationships, attributes etc.
+SQL injection should be explained with a demonstration (with a simple JS client).
+Concepts of database transaction, ACID properties, normal forms should be introduced with
+examples / live coding (creating a transaction, committing and rollback-ing).
```
remarcmij:
```suggestion
examples / live coding (creating a transaction, committing and rolling back).
```

Week3/sql-injection.js
```diff
@@ -0,0 +1,50 @@
...
+        console.error(error);
+    }
+    
+    for (i in results) {
+        console.log(results[i]);
+    }
+    connection.end();
+}
+
+queryDatabase();
```
remarcmij: You have two try/catch blocks in a single function. There should only be one. Note also that in JS we teach that `for...in` should not be used with arrays (or preferably not used in any circumstance). We teach to use `for...of` or the `.forEach()` array method instead:

```js
async function queryDatabase() {
  connection.connect();

  try {
    prompt.start();
    const result = await input(['number']);
    const inputNumber = result.number;

    // 1. Naive way of passing the parameter to the query
    //const select_query = `select * from students WHERE student_number =  ${input_number};`

    // 2. Escaping the parameter ( replacing the unwanted characters)
    //const select_query = `select * from students WHERE student_number =` + connection.escape(input_number);

    // 3. Using a question mark syntax to do the escaping (AKA prepared statements)
    const select_query = `select * from students WHERE student_number = ?`;

    console.log(select_query);
    const results = await execQuery(select_query, inputNumber);

    for (const result of results) {
      console.log(results);
    }
  } catch (error) {
    console.error(error);
  }

  connection.end();
}
```

Week3/README.md
unmeshvrije: I provided the link for detailed description for each normal form. And we discussed the forms in class. So I only summarized them here.

Week3/README.md
unmeshvrije: I used rollback-ing on purpose because rollback is the actual command in mysql which rolls the transaction back.

PR #201, opened 2018-12-31
Week3

Damla78:


Week3/app.js
```diff
@@ -0,0 +1,278 @@
...
+
+/**Category */
+async function listCategory() {
+  let queryWhere = " where id in (select categoryId from userscatalogs where userid = " + activeUserId + ")";
+  const categList = await getResult('categories', queryWhere);
+  return '' + JSON.stringify(categList);
+}
+async function createCategory(name) {
+  //uuidv4()
+  const id = await getLastId('categories');
```
lao: With the AUTO INCREMENT, you will not need this anymore. =)

Week3/app.js
```diff
@@ -0,0 +1,278 @@
...
+  let query = "insert into categories values (" + id + ", '" + name + "')";
+  await executeQuery(query);
+  //When category is created, relation of user & category is also created.
+  await createUsersCatalogs(activeUserId, id, true, true);
+  return name + ' ==> New category is created.';
+}
+async function updateCategory(id, newName) {
+  let query = "update categories set name = '" + newName + "' where id=" + id;
```
lao: Just a hint, try to escape this values ```newName``` and ```id``` so you won't need to worry about SQL Injection. =)

Week3/server.js
```diff
@@ -0,0 +1,155 @@
...
+const port = 3000;
+
+//categories
+app.get('/categories', async (req, res) => {
+  const msg = await todoMethodList.listCategory();
+  res.send(JSON.parse(msg));
+});
+//{"category": { "name": "Sport Lesson" }}
+app.post('/categories', async (req, res) => {
+  const msg = await todoMethodList.createCategory(JSON.parse(JSON.stringify(req.body.category)).name);
```
lao: I didnt understand the ```JSON.parse(JSON.stringify(req.body.category))```
seems to me that the result of it will be exactly the original value of ```req.body.category```. So you could simply just use ```req.body.category.name```. Also, try to do some input verification first. That would prevent potential errors. =)

image

PR #203, opened 2019-01-05
Week3

RimSebai:


Week3/create-data.js
```diff
@@ -0,0 +1,31 @@
...
+  password: 'hyfpassword',
+  database: 'new_world'
+});
+
+const execQuery = util.promisify(connection.query.bind(connection));
+
+async function createData() {
+  let sql = "CREATE DATABASE IF NOT EXISTS todo_data"
+  let use_sql = "USE todo_data"
+  let users = "CREATE TABLE IF NOT EXISTS Users(Id varchar(100) NOT NULL PRIMARY KEY UNIQUE, FirstName varchar(50) NOT NULL, LastName varchar(50))"
```
lao: Try to look into AUTOINCREMENT: https://dev.mysql.com/doc/refman/8.0/en/example-auto-increment.html

If you check the link, you see that with auto increment you don't need to provide an id as it is generated by MySQL itself.

Week3/database.js
```diff
@@ -0,0 +1,106 @@
...
+const getAll = async function (id) {
+  let sql = `select * from users a join toDosList b on a.Id = b.UserId join todoItems c on b.listId = c.listId where a.Id = ?`
+  try {
+    return await execQuery(sql, [id]);
+  } catch (error) {
+    connection.end();
+    return error;
+  }
+};
+const createUser = async function (id, firstName, lastName) {
```
lao: With the auto-increment, you won't need to pass an id here. In real-world cases, the client, ex.: web page or mobile app, would not know the id for a new user as this will be the responsibility of the server. In this case, our MySQL server will control that for us.

Week3/database.js
```diff
@@ -0,0 +1,106 @@
...
+  let sql = `insert into Users (id,firstName,lastName) values (?,?,?)`
+  try {
+    return await execQuery(sql, [id, firstName, lastName]);
+  } catch (error) {
+    connection.end();
+    return error;
+  }
+};
+const createList = async function (UserId, category, prioritize) {
```
lao: Just for consistency, I would change the name of this ```UserId``` to ```userId```. Tools like lint can help you with that: https://eslint.org/docs/rules/camelcase =)

Week3/server.js
```diff
@@ -0,0 +1,101 @@
...
+  }).catch(error => console.log(error.message));
+});
+
+app.get('/get-items/:id', function (req, res) {
+  db.getAll(req.params.id).then(result => {
+    res.send(result);
+  }).catch(error => console.log(error));
+});
+
+app.delete('/delete-user/:id', function (req, res) {
```
lao: When you make use of these methods ```delete()```, ```get()```, ```post()``` from express.js, you are binding the requests to HTTP METHODS. So for a cleaner API you could make the paths like this:

```javascript
app.get('/user/:id', function (req, res) {
  db.getAll(req.params.id).then(result => {
    res.send(result);
  }).catch(error => console.log(error));
});
app.delete('/user/:id', function (req,res) {let id = req.params.id;
  db.deleteUser(id).then(result => {
    res.send(result);
  }).catch(error => console.log(error.message));
});
```

And the client would request the same paths but with different HTTP METHODS. I added some extra links for more info too. =)

A link for HTTP METHODS:  https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods.
And here is a link for express.js methods doc: https://expressjs.com/en/api.html#app.delete.method

Week3/server.js
```diff
@@ -0,0 +1,101 @@
...
+  let listId = req.params.id;
+  let todo = req.body.todo;
+  let description = req.body.description;
+  let prioritize = req.body.prioritize;
+  db.createItem(listId, todo, description, prioritize).then(result => {
+    res.send(result);
+  }).catch(error => console.log(error.message));
+});
+
+app.get('/get-items/:id', function (req, res) {
```
lao: I got a little confused here. As I understood, the ```db.getAll``` method should expect the user id for getting the user info and items. But this endpoint ```'/get-items/:id'``` doesn't give me the complete understanding that I should send an userId for it. I would change the endpoint to this ```'/get-items/:userId'``` to avoid any confusion. =)

lao: And your value would be in ```req.params.userId``` for use in the function.

PR #206, opened 2019-01-05
week3

karamad10:


Week3/home-work/db.js
```diff
@@ -0,0 +1,40 @@
...
+   database: 'todoapp'
+});
+
+const execQuery = util.promisify(connection.query.bind(connection));
+
+async function createData() {
+   let sql = "CREATE DATABASE IF NOT EXISTS todo_data"
+   let use_sql = "USE todo_data"
+   let users = `CREATE TABLE IF NOT EXISTS
+       Users(Id varchar(100) NOT NULL PRIMARY KEY UNIQUE,
```
lao: It is a good idea to make use of AUTO_INCREMENT here to help you along the way: https://dev.mysql.com/doc/refman/8.0/en/example-auto-increment.html

Week3/home-work/app.js
```diff
@@ -0,0 +1,79 @@
+const bodyParser = require('body-parser');
+const express = require('express');
+const app = express();
+const queries = require('./query.js');
+
+app.use(bodyParser.json());
+
+
+
+app.get('/user/:id', function (req, res) {
```
lao: The client usually dont know the id of the next user it will create. That is why it is a good idea to have some AUTO INCREMENT in the user table too. =)

Week3/home-work/query.js
```diff
@@ -0,0 +1,75 @@
...
+const createUser = async function (id, firstName, lastName) {
+   let sql = `insert into Users (id,firstName,lastName) values (?,?,?)`
+   try {
+      return await execQuery(sql, [id, firstName, lastName]);
+   } catch (error) {
+      connection.end();
+      return error;
+   }
+};
+const createList = async function (UserId, category, prioritize) {
```
lao: I would keep the cammel case here. Check this comment I made before: https://github.com/HackYourFuture/databases/pull/203/files#r253290748

PR #209, opened 2019-01-30
Zerisenay Solomon DB week1 HW 

ZerSolo:


Week1/homework/src/1-countries_8_million.js
```diff
@@ -0,0 +1,9 @@
+'use-strict';
+
+const select_function = require('./select-query-function');
+
+const sql_query = `SELECT name FROM country where population > 8;`;
+const message = 'Countries with population more than 8 million.';
+const column = 'name';
+
+select_function(sql_query, message, column);
```
lao: How about instead of separating these answers in different files, you make use of some kind of prompt library https://github.com/flatiron/prompt#readme ?

Week1/homework/src/select-query-function.js
```diff
@@ -0,0 +1,29 @@
...
+const connection = mysql.createConnection({
+  host: 'localhost',
+  user: 'hyfuser',
+  password: 'hyfpassword',
+  database: 'world',
+});
+
+const execQuery = util.promisify(connection.query.bind(connection));
+
+async function select_function(select_query, consoleMessage, column) {
```
lao: What if did you change the inputs to column, table, whereCondition, and consoleMessage?

ZerSolo: Thank you Lukas for your feedback. I appreciate it a lot.
For the inputs , is not clear for me. Do I have to add table input ? and the where condition you mean the query? It would be nice if you can clarify it more for me. 
Thanks,

Week1/homework/src/1-countries_8_million.js
ZerSolo: I have made changes as requested. 
-made all queries in one file using the prompt module.

Week1/homework/src/select-query-function.js
```diff
@@ -14,16 +15,98 @@ const execQuery = util.promisify(connection.query.bind(connection));
...
     }
   } catch (error) {
     console.error(new Error(`this is error message : ${error}`));
   }
   connection.end();
 }
 
-module.exports = select_function;
\ No newline at end of file
+prompt.start();
```
lao: Very nice! One file, as entry point. =)

PR #210, opened 2019-01-30
class18-wk1-database

mohamad-monir:


fversnel:
Hi @mohamad-monir 
Looking at your table definitions like `europe_continent`, `Rotterdam_population` they're very specific. What would happen if you were to create a table called `continent` and add a column 'name' to it? Then you can add all the continents of the world to it.
The trick with database design is to think in abstract terms and not concrete. So not a table about Rotterdam, but instead a table about Cities for example.
Kind regards,
Frank

mohamad-monir:
Hi @fversnel 
Thank you for ur feedback .
It is a habit I picked from javascript naming convention ...you have to be descriptive ..
I pay more attention to it next time ...

PR #212, opened 2019-02-01
Yaacob databases week1 homework

SignosRk:
the homework is inside a folder called Homework.


fversnel:
Hi @SignosRk 
Homework looks nice, but take a good look at way you insert cities in the database. Have a good look at the insert syntax: https://www.dofactory.com/sql/insert and see if there is another way to insert cities without having to provide a city_id but use auto increment instead.
Cheers,
Frank

SignosRk:
@fversnel  Hi bro.. Thanks for the feedback but I would like you to explain me a bit better what do you really mean because we made this homework basically having as a guide another file due to the lack of time we had last week. And I am reading the documentation you sent me but I don't see how can I apply what you meant.

PR #215, opened 2019-02-08
Homework 2 - part one - part two - ERD Draw

SignosRk:
Task number 4 of part 1 is not completed, could not find a way to make the conditional statement, it only returns the true values, If I manage to find it before sunday then I will update my PR and let the mentors know. The rest of the code is written


fversnel:
Hi @SignosRk 
Please use prepared statements for your queries. If you don't know what those are the nodejs mysql docs might clarify things for you.

About the TODO database: think about what benefits it will bring you to have the 'Marked Completed', and 'Tagged' as separate tables. What can help you is to what kind of information you like to extract from the database to build the application (i.e. which queries you will be writing). Try to ask yourself why and if it is necessary for a single TODO item to have multiple marked completed entries.

PR #216, opened 2019-02-08
ZerSolo Database Week2

ZerSolo:


fversnel:
Hi @ZerSolo,

There's no need to create stored procedures. The assignment is to use prepared statements.

About the TODO database. Think about what the purpose is to use an `ENUM` type for tags and categories. What are the benefits? What are the downsides? What happens if the user wants to add a new tag? 
Also: Why can a single TODO have multiple TODO statuses?

ZerSolo:
hi, @fversnel  thank you for your feedback. 
According to your feedback, I have made corrections in my code. Actually, for the procedures, I asked first Unmesh  and he gave me green light.  Anyhow it is OK, at least It helped me to know the usage of stored procedures. 
please have a check and I would like to hear your feedback. It would be nice if you can give me the feedback for the ER diagram ASAP. 
because it will help me in this week's HW. 
Thanks in advance,

PR #217, opened 2019-02-08
added homework week2

Wael-Alhomsi:


fversnel:
Hi @Wael-Alhomsi 
The assignment says to use prepared statements to query the database. Check the nodejs mysql documentation on how to use them if you don't know what those are.
For the TODO database design: ask yourself which columns should be unique, which onces are the primary key and which ones are ordinary columns. Is it really necessary to make every column part of the key?
Good luck!

PR #218, opened 2019-02-09
Reza week2- New World- Parts 1 and 2

ranjbarreza:


fversnel:
Hi @ranjbarreza 
1. The assignment says to use prepared statements. Find out what those are. Perhaps the nodejs mysql documentation might clarify this.
2. Where's your TODO database design?

PR #221, opened 2019-02-10
Week2

soumaiamamo:


fversnel:
Hi @soumaiamamo 
For part 1 it states in the MAKEME.md that: " Note that you are expected to get the input from user and use prepared statements to write the queries.".

Your TODO database design looks good. Clean and simple :)

PR #223, opened 2019-02-16
added homework week3

Wael-Alhomsi:


fversnel:
Hi @Wael-Alhomsi 
Good job on the homework. Could you please format the long queries over multiple lines? Makes it much easier to read them. Like so:
```sql
SELECT *
FROM users
WHERE user_id = ?
AND ...
```
Cheers,
Frank

PR #230, opened 2019-03-12
Yaseir Alkhwalda - homework week 1

YaseirAlkhwalda:
update homework


Week1/select.js
```diff
@@ -0,0 +1,45 @@
...
+});
+
+connection.connect();
+
+const args = process.argv;
+args.splice(0, 2);
+const retrieve = argument => args.includes(argument);
+
+
+if (retrieve('1'))
```
zhaobn: You can also use `switch` here :)

zhaobn:
👍 👍 👍

Week1/select.js
YaseirAlkhwalda: thanks for your feedback ^_^ I updated my homework.

PR #231, opened 2019-03-14
Ammar Vehbi - Homework Week1

ammarvehbi:


Week1/homework/src/results.js
```diff
@@ -0,0 +1,181 @@
...
+      console.log(`
+        `);
+    }
+  );
+  connection.end();
+}
+
+function two() {
+  connection.connect();
+  connection.query('SELECT name FROM countries WHERE name LIKE "%land%" LIMIT 20', function(
```
zhaobn: You don't have to limit here. A small thing: could there be countries with the same name? How do you deal with that?

Week1/homework/src/results.js
```diff
@@ -0,0 +1,181 @@
...
+
+function four() {
+  connection.connect();
+  connection.query('SELECT name, region FROM countries WHERE region = "Europe" LIMIT 20', function(
+    error,
+    results
+  ) {
+    if (error) throw error.message;
+    console.log(`
+        4- Countries on the continent ‘Europe’:
```
zhaobn: Same thing - this is not exactly what you return, you cap your results at 20

zhaobn:
👍 👍 👍

Week1/homework/src/results.js
ammarvehbi: Dear @zhaobn ,
I used limit because I am fetching country data from an API. And if I don't limit, it will be many rows and will be hard to see all of the data.

Week1/homework/src/results.js
ammarvehbi: done :+1:

ammarvehbi:
Dear @zhaobn ,
Thanks for your review. The corrections could be seen in the last commit. :pray: :slightly_smiling_face:

PR #233, opened 2019-03-15
Utku / Homework Week 1

u61:


Week1/homework/create-database.js
```diff
@@ -0,0 +1,53 @@
...
+  user: 'hyfuser',
+  password: '7566',
+  database: 'userdb'
+});
+
+const execQuery = util.promisify(connection.query.bind(connection));
+
+async function seedDatabase() {
+  const CREATE_DATABASE_AND_LOGIN = `
+    CREATE DATABASE IF NOT EXISTS world; USE world`;
```
zhaobn: Or, you can take out the table creation code, put them in another place where you tell `mysql.js` to use `database: world`.

Week1/homework/create-database.js
```diff
@@ -0,0 +1,53 @@
...
+  const CREATE_COUNTRY_TABLE = `
+    CREATE TABLE IF NOT EXISTS country (
+      Name TEXT,
+      Population INT,
+      Surface_Area_km2 INT,
+      Continent TEXT
+    );`;
+
+  const CREATE_CITY_TABLE = `
+    CREATE TABLE IF NOT EXISTS city (
```
zhaobn: This schema is perhaps enough for the homework, but a better practice is to also give each row an ID. We will talk about in next lessons, pay attention to "key"s.

Week1/homework/exec_query.js
```diff
@@ -0,0 +1,34 @@
...
+
+const select_query = [
+  `SELECT name FROM country WHERE population >= 8000000;`,
+  `SELECT name FROM country WHERE name LIKE '%land%'`,
+  `SELECT name, population FROM city WHERE population BETWEEN 500000 AND 1000000`,
+  `SELECT name FROM country WHERE continent LIKE '%Europe%'`,
+  `SELECT name FROM country ORDER BY surface_area_km2 DESC`
+];
+
+for (let i of select_query) {
```
zhaobn: Extra challenge: can you make this function to accept an input from the user, say, a number, and then give the requested answer? For example, if I type 2, I'm expecting the second question from the homework.

Week1/homework/create-database.js
u61: But I especially used this syntax to put them together.
Is there anything wrong with it ?

Week1/homework/create-database.js
u61: IDs are added.

PR #246, opened 2019-03-22
ufukcalimli - Week2

ufukcalimli:


Week2/homework/app.js
```diff
@@ -0,0 +1,115 @@
...
+
+connection.connect();
+
+try {
+  switch (command) {
+    case 'capital':
+      const capitalOfCountry = `SELECT c.Name,ci.Name FROM country AS c 
+                                LEFT JOIN city AS ci
+                                ON c.Capital=ci.ID
+                                WHERE c.Name="${query}";`;
```
saragerion: Can you update the queries to use prepared statements? https://github.com/mysqljs/mysql

saragerion:
Few comments about the diagram: 
- User entity: username usermail userpassword -> the user part in the label is redundant
- Todo entity: due_date done are reduntant, what about a name?
- Task: what if I want to reference the same tag between multiple tasks? Or if I have more tags within the same task?

saragerion:
Can you add the package.json file to the homework?

PR #248, opened 2019-03-23
Week2 - fatih

kfth:


homework/week2/index.js
```diff
@@ -0,0 +1,94 @@
...
+      message: `Enter the ${question}:`,
+    });
+  }
+  let sql;
+  switch (question) {
+    case 'country':
+      sql = `SELECT city.name
+      FROM city, country
+      WHERE city.id=country.capital 
+      AND country.name = "${input.text}";
```
saragerion: This will work, but it's not very secure. 
What about using Prepared statements (as mentioned in the assignment)? Please check the npm package's documentation for how to achieve this.

kfth: thank you so much for your review. I've missed out "prepared statement " requirement in the homework. I've also forgotten to add instructions for the second and the third part of the homework, to cap it all i forgot to add/commit ER diagram file. I've fixed all those issues.
part 2: part2-trigger.sql
part 3: odo-erd.png

PR #254, opened 2019-03-29
Week3 Homework

ahmetbibi:
I tried to figure out it. I managed but there are still some errors. I will look again when find a time to look. Now it is working but includes some bugs.

### DIFF
--- a/Week3/homework/app/appModel.js
+++ b/Week3/homework/app/appModel.js
@@ -1,0 +1,159 @@
+'user strict';
+const sql = require('./db.js');
+
+//ToDo object constructor
+class Todo {
+  constructor(toDo) {
+    this.Name = toDo.Name;
+    this.ListID = toDo.ListID;
+  }
+
+  // todo functions
+
+  static createToDo(newToDo, result) {
+    sql.query('INSERT INTO toDos SET ?', newToDo, (err, res) => {
+      if (err) {
+        // console.log('error: ', err);
+        result(err, null);
+      } else {
+        // console.log(res.insertId);
+        result(null, res.insertId);
+      }
+    });
+  }
+
+  static getToDoById(toDoId, result) {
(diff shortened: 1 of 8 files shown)


Week3/homework/app/appModel.js
```diff
@@ -0,0 +1,159 @@
...
+        result(null, res);
+      }
+    });
+  }
+
+  static removeToDo(id, result) {
+    sql.query('DELETE FROM toDos WHERE id = ?', id, (err, res) => {
+      if (err) {
+        // console.log('error: ', err);
+        result(null, err);
```
ibudasov: Non-blocking:

I'm not quite sure if it's the correct order of params. Shouldn't be `err` first?

Week3/homework/app/appModel.js
```diff
@@ -0,0 +1,159 @@
...
+      } else {
+        result(null, res);
+      }
+    });
+  }
+}
+
+class Users {
```
ibudasov: Non-blocking

I would split Todo, List, Users to different files, for sake of SRP

Week3/homework/app/appRoutes.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+module.exports = app => {
+  const toDoApp = require('./appController');
+
+  // todoList Routes
+  // for a todo
+  app
+    .route('/users/:userId/:toDoListId/:toDoId')
```
ibudasov: Non-blocking comment: 

please have a look at the naming of URL in REST interfaces:
https://apiguide.readthedocs.io/en/latest/build_and_publish/use_RESTful_urls.html

Most probably it's something not quite clear at first glance, so I would be happy to explain it to you tomorrow or in Slack if you want.

I'm saying so because you'll need it for your interviews, and its gonna be important for other team members

PR #255, opened 2019-03-29
Ammar Vehbi - Homework Week3

ammarvehbi:
To run the app properly please follow the instructions in `README_Homework.md` file.


Week3/homework/README_Homework.md
```diff
@@ -0,0 +1,78 @@
...
+    - using proper url
+
+    - inputting necessary information if needed at body in postman
+
+## Functions:
+
+**Users Part:**
+
+    - To create a new user (POST):
+    `http://localhost:3001/ToDoApp/createNewUser`
```
ibudasov: **Non-blocking comment:** 

when you describe PUT/POST/PATCH requests - it would be great to specify also what kind of body parameters it expects: names, type of values (like string, int, bool), length, etc.

There is a tool for specifying that: https://swagger.io

**I'm saying this because of**, for instance, you'll be a backend developer, and you'll build an API. Then you'll have a lot of talks with frontend-guys, who will be asking you about how your API works, what does it expect and other stuff like this.

But if you use swagger - you'll say: here is swagger file, everything is described there. Useful.

Week3/homework/README_Homework.md
```diff
@@ -0,0 +1,78 @@
...
+    - To get all users (GET):
+    `http://localhost:3001/ToDoApp/getAllUsers`
+
+    To remove a user (DELETE):
+    `http://localhost:3001/ToDoApp/removeUser/:userId`
+
+**Todo Lists Part:**
+
+    - To create a new todo list (POST):
+    `http://localhost:3001/ToDoApp/:userId/createNewTodoList`
```
ibudasov: **Non-blocking comment:** 

please have a look at the naming of URL in REST interfaces:
https://apiguide.readthedocs.io/en/latest/build_and_publish/use_RESTful_urls.html

Most probably it's something not quite clear at first glance, so I would be happy to explain it to you tomorrow or in Slack if you want.

**I'm saying so because** you'll need it for your interviews, and its gonna be important for other team members

Week3/homework/ToDoApp_dump.sql
```diff
@@ -0,0 +1,113 @@
...
+
+--
+-- Table structure for table `todolists`
+--
+
+DROP TABLE IF EXISTS `todolists`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+ SET character_set_client = utf8mb4 ;
+CREATE TABLE `todolists` (
+  `User_Id` varchar(50) NOT NULL,
```
ibudasov: Why did you choose `varchar(50)`? In my experience, I see mostly `int` or `uuid`

Week3/homework/ToDoApp_dump.sql
```diff
@@ -0,0 +1,113 @@
...
+--
+
+DROP TABLE IF EXISTS `todos`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+ SET character_set_client = utf8mb4 ;
+CREATE TABLE `todos` (
+  `ToDoList_Id` varchar(50) NOT NULL,
+  `Todo_Id` varchar(50) NOT NULL,
+  `Todo_Name` varchar(100) NOT NULL,
+  `Done` enum('true','false') DEFAULT NULL,
```
ibudasov: **Non-blocking**

`enum` — 👌🏻

Don't you want to specify default value as one of `enum`'s values? Let's say `false`

Week3/homework/ToDoApp_dump.sql
```diff
@@ -0,0 +1,113 @@
...
+-- Table structure for table `users`
+--
+
+DROP TABLE IF EXISTS `users`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+ SET character_set_client = utf8mb4 ;
+CREATE TABLE `users` (
+  `User_Id` varchar(50) NOT NULL,
+  `Name` varchar(50) NOT NULL,
+  `Email` varchar(50) NOT NULL,
```
ibudasov: **Non-blocking**

Email can be longer than 50:
https://7php.com/the-maximum-length-limit-of-an-email-address-is-254-not-320/

Week3/homework/ToDoApp_dump.sql
ibudasov: Ah, I see, you use `nanoid()`

Week3/homework/src/actions/createNewTodo.js
```diff
@@ -0,0 +1,35 @@
...
+        '${req.params.todoListId}',
+        '${nanoid()}',
+        '${req.body.todo}',
+        '${req.body.done}',
+        STR_TO_DATE( '${req.body.dueDate}', '%d/%m/%Y %H:%i:%s'),
+        '${req.body.tag}'
+        )`
+    );
+    connection.query(
+      `SELECT * FROM todos WHERE ToDoList_Id = '${req.params.todoListId}'`,
```
ibudasov: Look, here is the problem.

JS has the event loop: https://flaviocopes.com/javascript-event-loop/
What can happen here: 
- if INSERT query takes too long, let's say, the DB is heavy loaded
- then JS will run SELECT query immediately, without waiting for INSERT query. 
- it will raise an error

One of the solutions - to use callbacks or promises. The goal is - to respect the event loop anyway

Week3/homework/src/actions/removeTodoList.js
```diff
@@ -0,0 +1,17 @@
+'use-strict';
+
+const connection = require('../databaseConfig');
+
+const removeTodoList = (req, res) => {
+  connection.query(`DELETE FROM todolists WHERE ToDoList_Id = '${req.params.todoListId}'`);
```
ibudasov: Non-blocking comment:

There is a good practice to use LIMIT in your queries: https://dba.stackexchange.com/questions/52079/does-using-limit-improve-the-performance-and-is-it-noticeable

Long story short:

- in case of select of only one row, you want to return the result once the row is found because you're looking for only one row, and it would be wise to stop the search when the one is found. However, if you select by PK - SQL server will know it and will stop the search itself - because of PK is unique
- in case of deletion - it's good for security reasons, and also saves you from occasional deleting of multiple rows.

Week3/homework/src/actions/updateTodoList.js
```diff
@@ -0,0 +1,30 @@
+'use-strict';
+
+const connection = require('../databaseConfig');
+
+const updateTodoList = (req, res) => {
+  if (req.body.todoListName === '' || req.body.todoListName === null) {
```
ibudasov: Non-blocking:

I see here and there you do validation as the first step in your code. This is good. And it's a repetitive action.
There is DRY rule: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

I'm saying so because of your colleagues will tell you this as well. Better be prepared.

Week3/homework/README_Homework.md
ammarvehbi: I took a look in the link you suggested and tried to update my code accordingly. :pray:

Week3/homework/src/actions/removeTodoList.js
ammarvehbi: Done :pray:

Week3/homework/ToDoApp_dump.sql
ammarvehbi: Yes, I read the article in the link it has to be 254 chars I think.

Week3/homework/src/actions/createNewTodo.js
ammarvehbi: Yes, I didn't take into consideration the event loop while I wrote the code because the database was not heavy. But I should have. I made the corrections for it. Thanks :pray:

Week3/homework/ToDoApp_dump.sql
ammarvehbi: Done :pray:

Week3/homework/README_Homework.md
ammarvehbi: Since my internet signal is very poor I couldn't look into deeply swagger, bu definitely I will. For now I updated my `README_Homework.md` file. And also I will try to push my swagger file ASAP. Thanks a lot @ibudasov

ammarvehbi:
Dear @ibudasov ,

I updated my code regarding your suggestions. It was a really constructive review. I bookmarked new sources :slightly_smiling_face: . 

As for the swagger, I will try to push my swagger file ASAP. 

You can see the updated version of the code from commit named `updates the code after review`. Thanks again :pray:

PR #256, opened 2019-03-30
Utku / Homework Week 3

u61:
- I have separated each required action to different files.
- index.js is the main file.
- package.json is available to install dependencies.
- Dump file is created.

### DIFF
--- a/Week3/homework/src/README.md
+++ b/Week3/homework/src/README.md
@@ -14,9 +14,14 @@ Each function must be tested by its own method and endpoint.
 
 - POST /username/password
 
+- username must be string(max. 20 char)
+- password must be string(max. 20 char)
+
 ### Create To-Do List
 
 - POST /username/password/listDescription
+
+- listDescription must be string(max. 50 char).
 
 ### Add Remainder To List
 
@@ -24,22 +29,35 @@ Small Note: First remainder is fixed. Second one is variable that you want to pu
 
 - PATCH /username/password/listDescription/remainder/remainder
 
+- remainder must be in 'CCYY-MM-DD hh:mm:ss' format.
+
 ### Create To-DO Item
 
 - POST /username/password/listDescription/toDoDescription
+
+- toDoDescription must be a string(max. 50 char).
 
 ### Mark Item As Completed
 
 - PATCH /username/password/listDescription/toDoDescription
 
+- toDoDescription must be a string(max. 50 char).
+
 ### Delete User
 
 - DELETE /username/password
+
+- username must be string(max. 20 char)
+- password must be string(max. 20 char)
 
 ### Delete To-Do List
 
 - DELETE /username/password/listDescription
 
+- listDescription must be string(max. 50 char).
+
 ### Delete To-Do Item
 
 - DELETE /username/password/listDescription/toDoDescription
+
+- toDoDescription must be a string(max. 50 char).
--- a/Week3/homework/src/README.md
+++ b/Week3/homework/src/README.md
@@ -1,0 +1,45 @@
+# To-Do REST API Manual
+
+You can create database with dump.sql file.
+
+You can trigger server from Week3 folder by 'node .' command.
+
+Then you must switch to localhost:3000 from POSTMAN
+
+## Supported Functions
+
+Each function must be tested by its own method and endpoint.
+
+### Create User
+
+- POST /username/password
+
+### Create To-Do List
+
+- POST /username/password/listDescription
+
+### Add Remainder To List
+
+Small Note: First remainder is fixed. Second one is variable that you want to put in.
+
+- PATCH /username/password/listDescription/remainder/remainder
+
+### Create To-DO Item
+
+- POST /username/password/listDescription/toDoDescription
+
+### Mark Item As Completed
+
+- PATCH /username/password/listDescription/toDoDescription
+
+### Delete User
+
+- DELETE /username/password
+
+### Delete To-Do List
+
+- DELETE /username/password/listDescription
+
+### Delete To-Do Item
+
+- DELETE /username/password/listDescription/toDoDescription
--- a/Week3/homework/src/addRemainder.js
+++ b/Week3/homework/src/addRemainder.js
@@ -3,13 +3,14 @@
 function addRemainder(connection, errorHandler, req, res) {
   const username = req.params.username;
   const password = req.params.password;
-  const listDescription = req.params.listDescription;
+  const listId = req.params.listId;
   const remainder = req.params.remainder;
-  const queryString1 = 'SELECT user_id FROM users WHERE user_name = ? AND user_password = ?;';
-  connection.query(queryString1, [username, password], (err, rows) => {
+  const retrieveUserIdQuery =
+    'SELECT user_id FROM users WHERE user_name = ? AND user_password = ?;';
+  connection.query(retrieveUserIdQuery, [username, password], (err, rows) => {
     errorHandler(err, res);
-    const queryString2 = 'UPDATE lists SET remainder = ? WHERE user_id = ? AND description = ?';
-    connection.query(queryString2, [remainder, rows[0].user_id, listDescription], err => {
+    const addRemainderQuery = 'UPDATE lists SET remainder = ? WHERE user_id = ? AND list_id = ?';
+    connection.query(addRemainderQuery, [remainder, rows[0].user_id, listId], err => {
       errorHandler(err, res);
     });
   });
--- a/Week3/homework/src/addRemainder.js
+++ b/Week3/homework/src/addRemainder.js
@@ -1,0 +1,19 @@
+'use strict';
+
+function addRemainder(connection, errorHandler, req, res) {
+  const username = req.params.username;
+  const password = req.params.password;
+  const listDescription = req.params.listDescription;
+  const remainder = req.params.remainder;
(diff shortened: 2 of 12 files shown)


Week3/homework/src/addRemainder.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+function addRemainder(connection, errorHandler, req, res) {
+  const username = req.params.username;
+  const password = req.params.password;
+  const listDescription = req.params.listDescription;
+  const remainder = req.params.remainder;
+  const queryString1 = 'SELECT user_id FROM users WHERE user_name = ? AND user_password = ?;';
```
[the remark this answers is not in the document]
ibudasov: same goes for `queryString2`

Week3/homework/src/addRemainder.js
```diff
@@ -0,0 +1,19 @@
...
+  connection.query(queryString1, [username, password], (err, rows) => {
+    errorHandler(err, res);
+    const queryString2 = 'UPDATE lists SET remainder = ? WHERE user_id = ? AND description = ?';
```
ibudasov: Seems like you're relying here on `description` in WHERE condition. 
I advise against such a thing, because of imagine the situation:
- you had list description
- the description has changed because of, let's say user edited it
- your query wouldn't be able to find the list to update anymore. 

And also, have a look - you use `user_id` already to identify the user, and this is a reliable approach. 
Please use primary key of  `lists` there

Week3/homework/src/createToDoItem.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+function createToDoItem(connection, errorHandler, req, res) {
+  const username = req.params.username;
+  const password = req.params.password;
+  const listDescription = req.params.listDescription;
+  const toDoDescription = req.params.toDoDescription;
+  const queryString1 = 'SELECT user_id FROM users WHERE user_name = ? AND user_password = ?';
```
ibudasov: **Non-blocking:**

I see here and there you retrieve user as the first step in your code. And it's a repetitive action.
There is DRY rule: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

**I'm saying so because** of your colleagues will tell you this as well. Better be prepared.

Week3/homework/src/createToDoList.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+function createToDoList(connection, errorHandler, req, res) {
+  const username = req.params.username;
+  const password = req.params.password;
```
ibudasov: **Non-blocking:**

In real life projects, you should never store passwords as a plaintext: https://security.stackexchange.com/questions/120540/why-shouldnt-i-store-passwords-in-plaintext

Remember SQL Injection attack? if you store your passwords as plaintext, once they are stolen - they can be used for other services, because ppl tend to use the same password in multiple places.

Week3/homework/src/deleteToDoItem.js
```diff
@@ -0,0 +1,23 @@
...
+  const password = req.params.password;
+  const listDescription = req.params.listDescription;
+  const toDoDescription = req.params.toDoDescription;
+  const queryString1 = 'SELECT user_id FROM users WHERE user_name = ? AND user_password = ?';
+  connection.query(queryString1, [username, password], (err, rows1) => {
+    errorHandler(err, res);
+    const queryString2 = 'SELECT list_id FROM lists WHERE user_id = ? AND description = ?;';
+    connection.query(queryString2, [rows1[0].user_id, listDescription], (err, rows2) => {
+      errorHandler(err, res);
+      const queryString3 = 'DELETE FROM items WHERE list_id = ? AND description = ?';
```
ibudasov: **Non-blocking comment:**

There is a good practice to use LIMIT in your queries: https://dba.stackexchange.com/questions/52079/does-using-limit-improve-the-performance-and-is-it-noticeable

Long story short: 
- in case of select of only one row, you want to return the result once the row is found because you're looking for only one row, and it would be wise to stop the search when the one is found. However, if you select by PK - SQL server will know it and will stop the search itself - because of PK is unique
- in case of deletion - it's good for security reasons, and also saves you from occasional deleting of multiple rows. 

Look at your query right know: if you've got 3 records with the same description in the same list - all 3 of them will be deleted.

Week3/homework/src/dump.sql
```diff
@@ -0,0 +1,105 @@
...
+--
+
+DROP TABLE IF EXISTS `items`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `items` (
+  `item_id` int(11) NOT NULL AUTO_INCREMENT,
+  `list_id` int(11) NOT NULL,
+  `description` longtext COLLATE utf8mb4_unicode_ci DEFAULT NULL,
+  `completion` tinyint(1) NOT NULL DEFAULT 0,
```
ibudasov: `tinyint` — 👌🏻

Week3/homework/src/dump.sql
```diff
@@ -0,0 +1,105 @@
...
+  PRIMARY KEY (`item_id`),
+  KEY `fk2` (`list_id`),
```
ibudasov: **Non-blocking:**

It would be a good thing to name foreign keys properly.

**Saying this because of** 
- when you see those keys on the diagram, and when there a lot of them - you can easily get lost without names
- then FK fails - it would be great to see it's name in the error message, it immediately gives you a nice clue about what went wrong

Week3/homework/src/dump.sql
```diff
@@ -0,0 +1,105 @@
...
+  CONSTRAINT `fk2` FOREIGN KEY (`list_id`) REFERENCES `lists` (`list_id`) ON DELETE CASCADE
```
ibudasov: Cool! You use cascade deletion instead writing code for it! Keep doing so! 🤙🏻

Week3/homework/src/dump.sql
```diff
@@ -0,0 +1,105 @@
...
+-- Table structure for table `lists`
+--
+
+DROP TABLE IF EXISTS `lists`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `lists` (
+  `list_id` int(11) NOT NULL AUTO_INCREMENT,
+  `user_id` int(11) NOT NULL,
+  `description` longtext COLLATE utf8mb4_unicode_ci DEFAULT NULL,
```
ibudasov: See the problem? 

- you use field with `longtext` in WHERE part of the query. It guarantees the query to be slow
- it is not unique, which means multiple rows can be deleted with the same query, as I told in the comment about LIMITs
- it is DEFAULT NULL, but it can not actually be, because of we rely on it

Week3/homework/src/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+Then you must switch to localhost:3000 from POSTMAN
+
+## Supported Functions
+
+Each function must be tested by its own method and endpoint.
+
+### Create User
+
+- POST /username/password
```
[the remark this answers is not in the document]
u61: I've checked that link and learned a lot regarding to naming.

Week3/homework/src/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Add Remainder To List
+
+Small Note: First remainder is fixed. Second one is variable that you want to put in.
+
+- PATCH /username/password/listDescription/remainder/remainder
+
+### Create To-DO Item
+
+- POST /username/password/listDescription/toDoDescription
```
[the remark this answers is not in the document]
u61: I've added additional explanations for each input value as type and length in README.md

Week3/homework/src/addRemainder.js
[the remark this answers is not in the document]
u61: Variable names are changed with regard to their functionality.

Week3/homework/src/addRemainder.js
u61: URI is changed and connected to user_id as PK.

Week3/homework/src/createToDoList.js
u61: I searched for this Igor but i couldn't find any data type for passwords. I came across with sha like syntax but I am not sure if it is proper one ?

Week3/homework/src/deleteToDoItem.js
u61: I've changed this URI and connected it to item_id as PK as well.

Week3/homework/src/dump.sql
u61: Actually i didn't intend to use tinyint but after I learned that if we use boolean as data types, it will be written as tinyint in dump file :)

Week3/homework/src/dump.sql
u61: Foreign key names are changed with meaningful ones.

Week3/homework/src/dump.sql
u61: I understand the logic. Thanks for that.
longtexts are changed with varchar(50).

u61:
Hi Igor,

I've tried to do some corrections especially the blocking one :).
It was beneficial for me to share your experience.

спасибо за все.

PR #269, opened 2019-06-05
add homework week1

Muhannad1989:


Week1/homework.js
```diff
@@ -0,0 +1,253 @@
+let mysql = require('mysql');
+
+// establish db
+let db = mysql.createConnection({
+  host: 'localhost',
+  user: 'root',
+  password: 'password',
+  database: 'world',
```
harditsingh: You might want to leave this line out. In case the database 'world' is not present in mysql, then this would give an error.

Week1/homework.js
```diff
@@ -0,0 +1,253 @@
...
+
+// drop database
+let drop_db = `DROP DATABASE IF EXISTS world`;
+db.query(drop_db, (error, result) => {
+  if (error) throw error;
+  // console.log(result);
+});
+
+// create database
+let create_db = `CREATE DATABASE IF NOT EXISTS world`;
```

Week1/homework.js
```diff
@@ -0,0 +1,253 @@
...
+db.query(create_db, (error, result) => {
+  if (error) throw error;
+  // console.log(result);
+  console.log('Database WORLD has been created');
+});
+
+let use_db = `USE world`;
```

Week1/homework.js
```diff
@@ -0,0 +1,253 @@
...
+});
+
+// connect database
+db.connect(err => {
+  if (err) throw err;
+  console.log('Mysql connected....');
+});
+
+// drop database
+let drop_db = `DROP DATABASE IF EXISTS world`;
```
harditsingh: You should leave this part out if you want to use world.sql for loading data because if you run world.sql before running this, this command would remove the database.

Muhannad1989:
thank you @harditsingh for your feedback i will try to skip all this issue's in the next homework.

PR #271, opened 2019-06-06
Akif_DBW1_Homework

Akif2709:


Week1/homework.js
```diff
@@ -0,0 +1,70 @@
+var mysql = require('mysql');
+var connection = mysql.createConnection({
+  multipleStatements: true,
+  host: 'localhost',
+  user: 'hyfuser',
+  password: 'hyfpassword',
+});
+
+connection.connect();
+const databaseWorld = 'CREATE DATABASE IF NOT EXISTS world';
```

Week1/homework.js
```diff
@@ -0,0 +1,70 @@
...
+const useDatabase = 'USE world';
```

PR #272, opened 2019-06-07
Orhan_Databases_Week1_Homework

OrhanGoc:
Orhan_Databases_Week1_Homework

### DIFF
--- a/Week1/homework.js
+++ b/Week1/homework.js
@@ -1,0 +1,113 @@
+const mysql = require('mysql');
+const connection = mysql.createConnection({
+  host: 'localhost',
+  user: 'hyfuser',
+  password: 'hyfpassword',
+});
+
+connection.connect();
+
+// creates the 'world' database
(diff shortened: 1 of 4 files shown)


Week1/homework.js
```diff
@@ -0,0 +1,113 @@
...
+
+// creates the 'world' database
+const createDatabase = `CREATE DATABASE IF NOT EXISTS world`;
+connection.query(createDatabase, error => {
+  if (error) throw error;
+  console.log('Database has been created');
+});
+
+// uses the database created
+connection.query(`USE world`, error => {
```

Week1/homework.js
```diff
@@ -0,0 +1,113 @@
...
+const connection = mysql.createConnection({
+  host: 'localhost',
+  user: 'hyfuser',
+  password: 'hyfpassword',
+});
+
+connection.connect();
+
+// creates the 'world' database
+const createDatabase = `CREATE DATABASE IF NOT EXISTS world`;
```

PR #276, opened 2019-06-10
upgrade of last few qus week1 homework

danlaudk:


danlaudk:
homework should
1. require joins
2. be easier to grade (be replicable by grader, and have nice output for grader)

PR #287, opened 2019-06-16
semih week2

sigma21:
week2

### DIFF
--- a/Week2/queryPackage.js
+++ b/Week2/queryPackage.js
@@ -1,0 +1,90 @@
+
+const languagesInACountry = `with langs as
+                            (select country.Name, country.Code, countrylanguage.Language
+                                from new_world.country join new_world.countrylanguage
+                                on country.Code = countrylanguage.CountryCode)
+                            select Language from langs where Name = ?`;
+
+                            
+const capitalCityQuery = `with result as
+                        (select city.name as Capital, country.name as Country
+                            from new_world.country join new_world.city
+                            on country.Capital=city.ID)
+                        select Capital from result where Country = ?`
+
+
+const numberOfCitiesLangSpoken = `with result as
+                                (select countrylanguage.Language, countrylanguage.CountryCode, city.Name
+                                from new_world.countrylanguage join new_world.city
+                                on countrylanguage.CountryCode = city.CountryCode)
+                                select * from result where Language = ?`
+
+
+const languagesInARegion = `with result as
+                    (select country.Name, country.Region, country.Code, countrylanguage.Language
+                        from new_world.country join new_world.countrylanguage
+                        on country.Code = countrylanguage.CountryCode)
+                    select Language from result where Region = ?`
+
+
+const officialLanguagesInRegion = `with result as
+                                (select country.Name, country.Region, country.Code, 
+                                    countrylanguage.Language, countrylanguage.IsOfficial
+                                    from new_world.country join new_world.countrylanguage
+                                    on country.Code = countrylanguage.CountryCode)
+                                select Name from result where Region like '%Europe%' and 
+                                Language like '%Dutch%' and IsOfficial like '%T'`
+
+
+
+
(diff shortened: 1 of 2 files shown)


Week2/queryPackage.js
```diff
@@ -0,0 +1,90 @@
...
+
+# with result as
+#     (select name,count(distinct language) from languages group by name),
+# create trigger langcheck after insert on languages for each row IF count(DISTINCT language) > 10
+#     THEN set lc_messages = 'alert!';
+#     END IF;
+
+delimiter $$
+CREATE TRIGGER date_trigger
+    AFTER INSERT
```
unmeshvrije: This trigger should be fired BEFORE insert.

Week2/queryPackage.js
```diff
@@ -0,0 +1,90 @@
...
+        ON languages
+            FOR EACH ROW
+            BEGIN
+                DECLARE message VARCHAR(100);
+#                 DECLARE project_start_date datetime ;
+#                 DECLARE project_end_date datetime ;
+                DECLARE language_count INTEGER ;
+#                 SET project_start_date = (select start_date from projects where proj_no=new.proj_no);
```
unmeshvrije: Remove the unnecessary comments. They are hard to parse while grading.

Week2/queryPackage.js
```diff
@@ -0,0 +1,90 @@
...
+
+
+# with result as
+#     (select name,count(distinct language) from languages group by name),
+# create trigger langcheck after insert on languages for each row IF count(DISTINCT language) > 10
+#     THEN set lc_messages = 'alert!';
+#     END IF;
+
+delimiter $$
+CREATE TRIGGER date_trigger
```
unmeshvrije: date_trigger is not a good name for this trigger.

Week2/queryPackage.js
```diff
@@ -0,0 +1,90 @@
...
+                 SET language_count = (select count(distinct language) from languages group by name);
+                IF language_count >= 10
+                THEN
+                    set message = 'Number of languages are now 10 or more.';
+                    SET lc_messages=message; SIGNAL SQLSTATE '45000';
+                END IF;
+            END $$
+delimiter ;
+
+drop trigger date_trigger;
```
unmeshvrije: You should not drop the trigger, because it will not be effective otherwise.

Week2/week2.js
```diff
@@ -0,0 +1,77 @@
...
+        for (r of results) {
+            console.log(`${r}`)
+        }                                  
+    } catch(error) {
+        console.error(error);
+    }
+    db.end();
+}
+
+
```
unmeshvrije: I think you intended to write a menu driven program (see herehttps://www.geeksforgeeks.org/menu-driven-program-using-switch-case-c/)  which runs a while loop and keeps asking users for input.
your program will terminate after giving you the answer of exactly one query.

Week2/queryPackage.js
```diff
@@ -0,0 +1,90 @@
...
+
+
+const languagesInARegion = `with result as
+                    (select country.Name, country.Region, country.Code, countrylanguage.Language
+                        from new_world.country join new_world.countrylanguage
+                        on country.Code = countrylanguage.CountryCode)
+                    select Language from result where Region = ?`
+
+
+const officialLanguagesInRegion = `with result as
```
unmeshvrije: The `with result as` syntax is recommended when you want to reuse it in the subqueries. 
In your case, the `result` subquery is only used once in a single query. So it is not necessary to use that alias. The query could be written as 
```
                    select Language from (select country.Name, country.Region, country.Code, countrylanguage.Language
                        from new_world.country join new_world.countrylanguage
                        on country.Code = countrylanguage.CountryCode) where Region = ?`
```

wouterkleijn:
Thanks @unmeshvrije Can you make the necessary changes @sigma21 ?

PR #291, opened 2019-06-21
Akif_DBW3_Homework

Akif2709:


danlaudk:
@Akif2709 please push package.json

Akif2709:
Hello @danlaudk, I added packagejson and corrected dump file. I hope it is as you wish.

danlaudk:
no foreign key from item to list, instead uses fk from iscomplete to both item and list, which _could_ be ok but entry from iscomplete is not removed when on-delete-item.
there's no data in the dump.

Akif2709:
I see the data in my Mysql when i open in my own computer. I dony know what is tge wrong. I dumped with the command which is written in the lesson documents. Also i added "on delete cascade" for isComplete foreign keys. I suppose it is working and delete whole line. What i need to do ?

PR #294, opened 2019-06-22
Muhannad db week3

Muhannad1989:


Muhannad1989:
please change the user and the password to be : 
user: hyfuser
password: hyfpassword

danlaudk:
users_todo and todolist_items tables are never used. api is able to access raw tables, but only via a delete function . So strictly the design has an incomplete implementation (nobody would be able to actually use this api productively). Major comment: one should not use  tablenames in the api . The ambition is to be lauded, and what is there works!

PR #295, opened 2019-06-22
I made my homework, needs some work

Kani103:


danlaudk:
overall, taking shape in right direction so that's good. should be fixed: should be foreign keys. insert :id shouldn't (won't) work with auto-increment. :id shouldn't be in all routes  (see my (dan) post in slack to #class ).

danlaudk:
:id should not be in POST routes. GET requests are not supported. Otherwise design is good.  Please include two foreign keys. Please submit a working app *and a database dump* file (eg dump.sql) that is created AFTER you have made successful Post request (and i will run a GET against it to test)

wouterkleijn:
Thanks for reviewing @danlaudk Please have a look @Kani103

wouterkleijn:
Thanks @Kani103 , what about the changes @danlaudk - any further feedback?

PR #297, opened 2019-06-22
Mohammad-Darwesh-homework-db-week3

mohammadnaefaldarwesh:


Week3/homework/update.js
```diff
@@ -0,0 +1,47 @@
...
+const { errorHandler, userError, listError, todoError } = require('./errorHandler');
+const constraints = {
+  from: {
+    email: true,
+  },
+};
+
+const user = async (req, res) => {
+  const body = req.body;
+  isEmpty(body) ||
```
danlaudk: why is this isEmpty(body) useful?

danlaudk:
good error handling and messages. things to consider: default values  - eg. done = "false" , eg on update = prior value. good job!

Week3/homework/todoApp.sql
```diff
@@ -0,0 +1,121 @@
...
+CREATE TABLE todos
+(
+  `todolist_id` varchar
+(50) NOT NULL,
+  `todo_id` varchar
+(50) NOT NULL,
+  `todo_name` varchar
+(100) NOT NULL,
+  `done` enum
+('true','false') DEFAULT 'false',
```
danlaudk: why enum and not boolean?

Week3/homework/update.js
mohammadnaefaldarwesh: Hello @danlaudk ,
Thank you for your feed back .
isEmpty(body)  is not useful after i validated the user and the email but i did not notice that ...In the first i validated only the body and after few days i decide to validate the user and all required fields.... On that time i did not notice that validate the body will be not necessary any more  ....So it is my mistake...I should check my code carefully before submit my homework.
Best regards,

Week3/homework/todoApp.sql
mohammadnaefaldarwesh: Hello @danlaudk ,
I used enum because i want the value to be  `only` true or false ......I think if i used boolean 1 and 0 will considered also as true and false  .....If do you think boolean will be better in this situation please can you explain to me why ? 
Best regards,

danlaudk: you're right about 1/0 . i'd question why you don't want to allow 1/0 (ie only true false). 

both js and db natively support boolean : takes less space, is standard way to encode binary value. Unless you expect to add more values to the enum in the future (a 3rd value called "almost done", and then you should probably code that already) ... So mostly it looks really weird (because it's non standard), making the person reading your code scratch their head and ask "why not boolean"?

PR #305, opened 2019-07-23
hamza-week1

lizozo:


Week1/Node.js
```diff
@@ -0,0 +1,62 @@
...
+  host: 'localhost',
+  user: 'hyfuser',
+  password: 'hyfpassword',
+});
+connection.connect();
+
+let queries = [
+  'CREATE DATABASE world',
+  'USE world',
+  'CREATE TABLE country (name VARCHAR(56), Continent ENUM("Asia","Europe","North America","Africa","Oceania","Antarctica","South America"), Region VARCHAR(50), SurfaceArea FLOAT(10,2),' +
```
gajduk: nice use of enum 👍

Week1/Node.js
```diff
@@ -0,0 +1,62 @@
...
+    'IndepYear SMALLINT(5), Population INT(11), LifeExpectancy FLOAT(3,1), GNP FLOAT(10,1), GNPOld FLOAT(10,1), LocalName VARCHAR (74),' +
+    'GovernmentForm VARCHAR(50), HeadOfState VARCHAR(30), Capital VARCHAR(30))',
+  'CREATE TABLE city (ID INT, Name VARCHAR(50), CountryCode VARCHAR(10), District VARCHAR(50), Population FLOAT(8,1))',
```
gajduk: Can a population have a decimal value? 🤔

lizozo: Hi @gajduk !! Thanks for the feedback! I dont think  population can have a decimal value, unless it is presented in (X1000 or K). if 1.200.500 population is asked to be presented in Ks, 1200,5K, it can be decimal. but i dont have such restiriction. so I set the population type as integer which is a whole number type.

lizozo: oooh sorry @gajduk !! I thought it is the population in country table. Yes, you are right. The population in city table should be integer. should i correct it and  make a pull request?

PR #306, opened 2019-07-23
Suheyb - Week1

suh3yb:


Week1/node.js
```diff
@@ -0,0 +1,90 @@
...
+
+// 1. What are the names of the countries with population greater than 8 million
+// 2. What are the names of the countries that have “land” in their names ?
+// 3. What are the names of the cities with population in between 500,000 and 1 million ?
+// 4. What are the names of all the countries on the continent ‘Europe’ ?
+// 5. List all the countries in the descending order based on their surface areas.
+
+const queriesForSearch = [
+  "SELECT name AS 'More Than 8 Million Population' FROM countries WHERE population > 8000000",
+  "SELECT name FROM countries WHERE name LIKE '%land'",
```
gajduk: `%land` will only match countries where the name **ends** in *land*. The assignments says to find all countries that have *land* in their names which means it can be at any position in the name, not only on the end. Then again I can not think of any country that has *land* in the name, but it is not at the end.

Week1/node.js
```diff
@@ -0,0 +1,90 @@
...
+
+// 6. What are the names of all the cities in the Netherlands?
+// 7. What's the population of Rotterdam?
+// 8. What's the top 10 countries based on surface area?
+// 9. What's the top 10 cities with the highest population?
+// 10. What's the population of the world ?
+
+const optionalQueriesForSearch = [
+  "SELECT name AS 'Cities in The Netherlands' FROM cities WHERE country_code = 31",
+  "SELECT population AS 'Population of Rotterdam' FROM cities WHERE name = 'Rotterdam'",
```
gajduk: like that you renamed the result 👍

Week1/node.js
suh3yb: Oh, I did not realize that. In my code I put Switzerland and Netherlands into the database. `%land` only matches with Switzerland and not Netherlands. I will correct this, thanks.

suh3yb:
Thank you very much @gajduk for your feedback.
I have changed `%land` to `%land%`.

PR #315, opened 2019-07-26
aalhommada-database-week1

aalhommada:


Week1/node.js
```diff
@@ -0,0 +1,184 @@
...
+  user: 'hyfuser',
+  password: 'hyfpassword',
+  database: 'world',
+});
+
+database.connect();
+
+// create database world
+
+database.query(`CREATE DATABASE IF NOT EXISTS world`, (error, result) => {
```
gajduk: if not exists 👍

Week1/node.js
```diff
@@ -0,0 +1,184 @@
...
+// use world database
+database.query(`USE world`, (error, result) => {
+  if (error) throw error;
+  console.log('change to WORLD database');
+});
+
+//========== create tables : country , city , and insert data in it =================
+
+const table_of_country =
+  'CREATE TABLE IF NOT EXISTS country (name VARCHAR(50), continent VARCHAR(50), region VARCHAR(50), surface_area FLOAT, indep_year INT, population INT, life_expectancy, GNP FLOAT, GNPold FLOAT, local_name VARCHAR(50), government_form VARCHAR(50), head_of_state VARCHAR(50),capital VARCHAR(50))';
```
gajduk: what is the type of `life_expectancy` ?

Week1/node.js
```diff
@@ -0,0 +1,184 @@
...
+var table_of_city =
+  'CREATE TABLE IF NOT EXISTS city (id INT, name VARCHAR(50),country_code VARCHAR(10), district VARCHAR(50), population INT)';
+
+database.query(table_of_city, (error, results) => {
+  if (error) throw error;
+  console.log('City table is created');
+});
+
+var insert_countries = [
+  "INSERT INTO country (name, continent,surface_area, population) VALUES ('Angola','Africa',1246700,12878000)",
```
gajduk: did you know that you can have multiple values in a single query
``` 
"INSERT INTO country (name,     continent,  surface_area,  population) VALUES "+
"                    ('Angola', 'Africa',   1246700,       12878000),   "+
"                    ('Syria',  'Asia',     185180,        18269868),   "+ // and so forth
```

Week1/node.js
```diff
@@ -0,0 +1,184 @@
...
+  if (error) throw error;
+
+  console.log(`\n Countries with population more than 8 million:`);
+  result.forEach(country => {
+    console.log(`${country.Name} , ${country.Population}`);
+  });
+});
+
+// Names of the countries that have “land” in their names ?
+database.query(`SELECT * FROM country WHERE Name LIKE '%land%' `, (error, result) => {
```
gajduk: Using `select *` is not recommended in real systems. Be specific with the result and only return the columns that you need, in this case the name.

Week1/node.js
```diff
@@ -0,0 +1,184 @@
...
+
+  console.log(`\n Countries in Europe continent:`);
+  result.forEach(country => {
+    console.log(country.Name);
+  });
+});
+
+// Countries in descending order, based on their surface areas.
+
+database.query(`SELECT * FROM country ORDER BY SurfaceArea DESC`, (error, result) => {
```
gajduk: I believe the column name is `surface_area`

Week1/node.js
```diff
@@ -0,0 +1,184 @@
...
+  if (error) throw error;
+
+  console.log(`\n Top 10 countries based on surface area:`);
+  result.forEach(country => {
+    console.log(country.Name);
+  });
+});
+
+// top 10 cities with the highest population?
+database.query('SELECT FROM city ORDER BY population DESC LIMIT 10', (error, result) => {
```
gajduk: select what??

Week1/node.js
aalhommada: indeed i did not know , but it is implemented now , thanks a lot.

Week1/node.js
aalhommada: Thanks

Week1/node.js
aalhommada: FLOAT. i add it , thanks

Week1/node.js
aalhommada: I did not know,  i implement it

Week1/node.js
aalhommada: Thanks , i make it name

Week1/node.js
aalhommada: That is right , thanks

Week1/node.js
aalhommada: Thanks for mention that , it is corrected now

PR #316, opened 2019-07-27
Queries for week1-irfan

irfankaradeniz:


Week1/queries.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+  console.log('Connected.');
+});
+
+const queries = {
+  createDatabase: `CREATE DATABASE IF NOT EXISTS world;`,
+
+  useDatabase: `USE worldData`,
+
+  createTableCountry: `CREATE TABLE IF NOT EXISTS country (code CHAR(3), Name CHAR(52), Continent CHAR (30), Region CHAR(26),SurfaceArea FLOAT(10,2),Indepyear SMALLINT(6),Population INT(11),LifeExpectancy FLOAT(3.1),GNP FLOAT(10.2),LocalName CHAR(45),GovernmentForm CHAR(45),HeadOfState CHAR(60),Capital INT(11),Code2 CHAR(2),PRIMARY KEY(Code));`,
```
gajduk: I see you are using CHAR a lot. This is not wrong per se but it is good to understand the difference

https://stackoverflow.com/questions/1885630/whats-the-difference-between-varchar-and-char

Week1/queries.js
```diff
@@ -0,0 +1,98 @@
...
+
+  createTableCity: `CREATE TABLE IF NOT EXISTS city(ID INT(11) AUTO_INCREMENT,Name CHAR(35),CountryCode CHAR(3),District CHAR(20),Population INT(11),PRIMARY KEY (ID),KEY CountryCode (CountryCode))`,
```
gajduk: Nice that you defined primary keys.
What type of key is CountryCode? What table and column does it refer to?

Week1/queries.js
```diff
@@ -0,0 +1,98 @@
...
+    if (error) {
+      throw error;
+    }
+    console.log('Query successful...');
+  });
+});
+
+//Queries
+// i.
+const country_pop8mil = `SELECT * FROM country WHERE Population > 8000000`;
```
gajduk: using `select *` in real system is not recommended. Always only retrieve things that you actually intend to use https://stackoverflow.com/questions/3639861/why-is-select-considered-harmful

Week1/queries.js
```diff
@@ -0,0 +1,98 @@
...
+    throw error;
+  }
+  console.log(`Countries with “land” in their names:`);
+  data.forEach(element => {
+    console.log(element.Name);
+  });
+});
+
+// iii.
+const population_between = `SELECT * FROM city WHERE Population BETWEEN 500000 AND 1000000 `;
```
gajduk: using between 👍

Week1/queries.js
irfankaradeniz: I read it and understood. But can you give me an example of it in my queries? For example for the first query.. How can I do that instead of using select * ?

Week1/queries.js
irfankaradeniz: Actually the query was in the real sql file. Now I changed chars into varchars. Thanks!

Week1/queries.js
irfankaradeniz: I got that. I made a reference to countries(code).

irfankaradeniz:
Thanks for your feedback!

Week1/queries.js
irfankaradeniz: Okai, I think I got that. I committed updated version with the changes of 'select *'.

Week1/queries.js
```diff
@@ -18,9 +18,9 @@ const queries = {
 
   useDatabase: `USE worldData`,
 
-  createTableCountry: `CREATE TABLE IF NOT EXISTS country (code CHAR(3), Name CHAR(52), Continent CHAR (30), Region CHAR(26),SurfaceArea FLOAT(10,2),Indepyear SMALLINT(6),Population INT(11),LifeExpectancy FLOAT(3.1),GNP FLOAT(10.2),LocalName CHAR(45),GovernmentForm CHAR(45),HeadOfState CHAR(60),Capital INT(11),Code2 CHAR(2),PRIMARY KEY(Code));`,
+  createTableCountry: `CREATE TABLE IF NOT EXISTS country (code char(3), Name varchar(52), Continent varchar (30), Region varchar(26),SurfaceArea FLOAT(10,2),Indepyear SMALLINT(6),Population INT(11),LifeExpectancy FLOAT(3.1),GNP FLOAT(10.2),LocalName varchar(45),GovernmentForm varchar(45),HeadOfState varchar(60),Capital INT(11),Code2 varchar(2),PRIMARY KEY(Code));`,
 
-  createTableCity: `CREATE TABLE IF NOT EXISTS city(ID INT(11) AUTO_INCREMENT,Name CHAR(35),CountryCode CHAR(3),District CHAR(20),Population INT(11),PRIMARY KEY (ID),KEY CountryCode (CountryCode))`,
+  createTableCity: `CREATE TABLE IF NOT EXISTS city(ID INT(11) AUTO_INCREMENT,Name varchar(35),CountryCode varchar(3),District varchar(20),Population INT(11),PRIMARY KEY (ID),FOREIGN KEY(country_code) REFERENCES countries(code))`,
```
gajduk: correct 👍

Week1/queries.js
```diff
@@ -34,9 +34,9 @@ Object.keys(queries).forEach(function(item) {
   });
 });
 
-//Queries
+// Queries
 // i.
-const country_pop8mil = `SELECT * FROM country WHERE Population > 8000000`;
+const country_pop8mil = `SELECT Name,Population FROM country WHERE Population > 8000000`;
```
gajduk: 👍

PR #319, opened 2019-07-31
aalhommada database week2

aalhommada:


fedefu:
Thanks for the feedback @ufukcalimli! can you please implement the feedback @aalhommada ?

PR #328, opened 2019-08-03
Database week2 irfan

irfankaradeniz:


Week2/nodeapp.js
```diff
@@ -0,0 +1,91 @@
...
+    await execQuery(`use world`);
+
+    if (userInput === '1') {
+      console.log('Enter the country name:');
+      const result = await input(['userInput']);
+      const country = result.userInput;
+      const data = await execQuery(
+        `with result as 
+        (select country.name as Country, city.name as Capital
+        from world.country join world.city on country.Capital=city.ID)
```
gajduk: Since you already defined the database in the connection it is not necessary  to do it here.  In fact it is better not to since often times you will need to run the same application but with different databases.

Week2/nodeapp.js
```diff
@@ -0,0 +1,91 @@
...
+        select Country, Capital from result where Country like '%${country}%'`,
```
gajduk: What would happen if the user enters the following query:
Spain; DROP TABLE country
You should be using prepared statements to guard your app against a hacker attack.
You can find an example here https://github.com/mysqljs/mysql#preparing-queries

Week2/nodeapp.js
```diff
@@ -0,0 +1,91 @@
...
+    } else if (userInput === '3') {
+      console.log('Enter the language name:');
+      const result3 = await input(['userInput']);
+      const language = result3.userInput;
+      const dataLang = await execQuery(
+        `with result as
+        (select countrylanguage.Language, countrylanguage.CountryCode, city.Name
+        from world.countrylanguage join world.city
+        on countrylanguage.CountryCode = city.CountryCode)
+        select language, count(Name) as NumberOfCities from result where language like '%${language}%'`,
```
gajduk: When you use an aggregated query (such as Count, Sum or Avg) then you can not have non-aggregated columns in the select. How can you fix this?

Week2/nodeapp.js
```diff
@@ -0,0 +1,91 @@
...
+        from world.country join world.countrylanguage on country.Code = countrylanguage.CountryCode)
+        select Name from result where Region like '%${regionInput}%' and Language like '%${langInput}%' and IsOfficial like '%T'`,
+      );
+      if (dataQuery4.length === 0) {
+        console.log('False');
+      } else {
+        console.log(dataQuery4);
+      }
+    } else if (userInput === '5') {
+      const dataQuery5 = await execQuery(`select country.continent as Continent, count(countrylanguage.language) as NumberOfLanguages
```
gajduk: Suppose you have a small continent with only two countries that speak the same language. On this continent only 1 language is spoken. However your query would return a different result. How can you fix this?

fedefu:
Thanks for the feedback @gajduk. Can you please implement the feedback @irfankaradeniz? Cheers

PR #332, opened 2019-08-09
hamza-week3

lizozo:


Week3/index.js
```diff
@@ -0,0 +1,249 @@
...
+    const user_name = request.params.user_name;
+    const list_name = request.params.list_name;
+    if (listIdValue == '' || userIdValue == '') {
+      response.status(400).send('user name or list name does not exist');
+    } else {
+      const { todo_desc } = request.body;
+      const userIdValue = await idToValue('user_id', 'user', 'user_name', user_name);
+      const listIdValue = await idToValue('list_id', 'todo_list', 'list_name', list_name);
+      const sql = createQuery([
+        'UPDATE `todo_item` SET `is_done` = 1 WHERE `user_id` = ? AND `list_id` = ? AND todo_desc = ?',
```
gajduk: I am curious why here you are not checking if userId and listId are valid? It seems to me that it would be easy to forget that the function idToValue returns an empty string when the id is not found? This could lead to some nasty bugs. Best not to leave such things to chance. How could you ensure that no one makes a mistake by being careless like that?

Week3/todo.sql
```diff
@@ -0,0 +1,115 @@
...
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+ SET character_set_client = utf8mb4 ;
+CREATE TABLE `todo_item` (
+  `todo_id` int(11) NOT NULL AUTO_INCREMENT,
+  `todo_desc` varchar(45) NOT NULL,
+  `is_done` tinyint(1) NOT NULL DEFAULT '0',
+  `list_id` int(11) NOT NULL,
+  `user_id` int(11) NOT NULL,
+  PRIMARY KEY (`todo_id`),
+  UNIQUE KEY `id_UNIQUE` (`todo_id`),
```
gajduk: primary keys are unique by default no need to specify unique explicitly again

Week3/todo.sql
```diff
@@ -0,0 +1,115 @@
...
+
+DROP TABLE IF EXISTS `todo_item`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+ SET character_set_client = utf8mb4 ;
+CREATE TABLE `todo_item` (
+  `todo_id` int(11) NOT NULL AUTO_INCREMENT,
+  `todo_desc` varchar(45) NOT NULL,
+  `is_done` tinyint(1) NOT NULL DEFAULT '0',
+  `list_id` int(11) NOT NULL,
+  `user_id` int(11) NOT NULL,
```
gajduk: seems to me that todo_list already has a reference to user so storing another reference to user in the todo_item violates 3rd normal form. Unless I am misunderstanding something

lizozo:
hello @gajduk thanks for your valuable feedback. you are right that user_id column in todo_item is redundant [here](https://github.com/HackYourFuture/databases/pull/332#discussion_r313056174).

[here](https://github.com/HackYourFuture/databases/pull/332#discussion_r313053766) it seems that the variables that i am using in if blocked are declared after the if block. I just need to move them before the if block. For the returning a empty string, you are right it is a vulnerable method to check.

Thanks again Andrej. you are great.

gajduk:
Thank you. To go back to the idToValue function compare the following two implementations
```
function idToValue(column1, table', column2, value) {
   return executeQuery(`select ....`).getResult();
} 
app.get('/user/:name', (req, res) => {
   try {
     let userid = idToValue('user_id', 'user', 'user_name', name);
     if ( userid == '' ) res.send(400).end('no such user);
     else {
        let user = getUserDetails(user_id);
        res.send(200).end(user);
     }
   catch ( e ) {
       res.send(400).end(e.message);
    }
})
```
versus
```
function idToValue(column1, table', column2, value) {
   let res = executeQuery(`select ....`).getResult();
   if ( res == '' ) throw new Error('no such user');       
} 
app.get('/user/:name', (req, res) => {
   try {
     let userid = idToValue('user_id', 'user', 'user_name', name);
     let user = getUserDetails(user_id);
     res.send(200).end(user);
   } catch ( e ) {
       res.send(400).end(e.message);
    }
})
```

PR #337, opened 2019-08-10
Week3

hanialkudsi:


Week3/ToDoList.sql
```diff
@@ -0,0 +1,79 @@
...
+-- Table `todos`.`todo_item`
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS `todolist`.`todo_item` (
+  `item_id` INT UNSIGNED NOT NULL AUTO_INCREMENT,
+  `list_id` INT NOT NULL,
+  `title` VARCHAR(45) NOT NULL,
+  `description` VARCHAR(45) NULL,
+  `completed` TINYINT NULL DEFAULT 0,
+  `created_at` DATETIME NULL DEFAULT CURRENT_TIMESTAMP,
+  PRIMARY KEY (`item_id`),
```
gajduk: correct use of primary and foreign keys 👍

Week3/routs.js
```diff
@@ -0,0 +1,98 @@
...
+  database: 'toDoList',
+});
+
+module.exports = function(app) {
+  app.get('/server', (req, res) => res.send('pong'));
+
+  app.post('/todoList/createOne', (req, res) => {
+    const newList = req.body.list_name;
+    const userId = req.body.user_id;
+    const addNewList = `insert into todo_list (user_id,list_name) values ('${userId}','${newList}')`;
```
gajduk: You are making two errors here:
1) You are not using prepared statements. A malicious entity could send a request where the userId is 'Andrej; drop table users' and delete your entire users table
2) according to the sql above the user_id is auto increment. So it should not be set manually, because that defeats the purpose of auto_increment not to mention nothing guarantees uniqueness in the ids that clients send.

Week3/server.js
```diff
@@ -0,0 +1,26 @@
+const express = require('express');
+const app = express();
```
gajduk: please do not mix requires with other code 😢

Week3/server.js
```diff
@@ -0,0 +1,26 @@
...
+  password: 'hyfpassword',
+  database: 'toDoList',
+});
+
+connection.connect(function(err) {
+  if (err) throw err;
+  console.log('database connected');
+});
+
+require('./routs')(app);
```
gajduk: require should go on top 🙃

Week2/srever.js
```diff
@@ -71,7 +71,7 @@ app.get('/countries/:region', (req, res) => {
 // get numbers of languages in a all continents
 // http://localhost:5055/continent/languages
 app.get('/continent/languages', (req, res) => {
-  var part1_5 = `select DISTINCT(continent),count(language) from country as co
+  var part1_5 = `select DISTINCT continent, count (DISTINCT language) from country as co
```
gajduk: correct

PR #340, opened 2019-08-16
aalhommada-database-week3

aalhommada:


Week3/week3-homework/actions/action.js
```diff
@@ -0,0 +1,104 @@
...
+});
+
+
+const execQuery = util.promisify(connection.query.bind(connection));
+connection.connect();
+
+async function showTable(req, res) {
+  const tableName = req.params.table;
+  try {
+    const showTableData = await execQuery(`SELECT * FROM ` + (tableName));
```
gajduk: suppose I call the following endpoint
`http://lcoalhost:3000/users; DROP TABLE users`
What would happen?

Week3/week3-homework/todo-app.js
```diff
@@ -0,0 +1,40 @@
...
+const bodyParser = require('body-parser');
+const app = express();
+app.use(bodyParser.json());
+
+
+app.get('/:table', (req, res) => {
+  showTable(req, res)
+});
+
+app.post('/add_todo_item', (req, res) => {
```
gajduk: This path does not conform to REST. The path should be simply `/todo_item`, the method type `post `already communicated the intention to create an object. Same for delete it should be simply
`app.delete('/todo_item/:id' ...`
The method is `delete `so there is no need to have delete in the path as well.

PR #344, opened 2019-09-09
mehmet bozdemir databases week 1 homework

mehmet-bozdemir:


Week1/homework/createTable.js
```diff
@@ -0,0 +1,22 @@
...
+  user     : 'hyfuser',
+  password : 'hyfpassword',
+  database : 'myWorld'
+});
+
+connection.connect(function(err) {
+    if (err) throw err;
+    console.log("Connected!");
+    var tables = [
+        "CREATE TABLE country (Country_ID int, Country_Name char(52) ,Continent enum('Asia','Europe','North America','Africa','Oceania','Antarctica','South America') ,Region char(50),SurfaceArea float,Population int,LifeExpectancy float, GNP float ,GNPOld float, LocalName char(50),GovernmentForm char(50),HeadOfState char(50),Capital char(50))",
```
mogaleaf: This is good but I would recommend you to use varchar instead of char when you have String with different size as it will take less space in the db.

Week1/homework/queries.js
```diff
@@ -0,0 +1,114 @@
+var mysql      = require('mysql');
```
mogaleaf: You did not provide any INSERT in your homework, you need some (and I need some) in order to test this queries :)

Week1/homework/queries.js
```diff
@@ -0,0 +1,114 @@
...
+  database : 'myworld'
+});
+
+connection.connect(function(err) {
+    if (err) throw err;
+    console.log('db connected');
+  });
+
+// Question 4i: What are the names of the countries with population greater than 8 million
+let result4i = 'SELECT Country_Name FROM Country WHERE Population > 8000000;';
```
mogaleaf: You don't need the ';' at the end when you are creating the query with nodejs.

Week1/homework/queries.js
```diff
@@ -0,0 +1,114 @@
...
+
+connection.query(result4ii, function(error, data) {
+  if (error) {
+    throw error;
+  }
+  console.log(data);
+});
+
+// Question 4iii: What are the names of the cities with population in between 500,000 and 1 million ?
+let result4iii = "SELECT City_Name FROM City WHERE Population BETWEEN 500.000 AND 1000000;";
```
mogaleaf: Careful 500.000 will be considered as float so basically 500 and not as Int 500000.

mehmet-bozdemir:
Dear Cecile, thank you for the feedback, I am gonna implement them, there was a small misunderstanding about using which database for the questions 4 and 5 now I am clear after your correction and also Unmesh's clearance on slack this morning.

mehmet-bozdemir:
Dear Cecile, I implemented your feedback as required, I already know that there is no need to use quotes when inserting values into tables but since I am lazy I just copied rows from actual world database in order to populate my own database. 
Again thank you for your guidance.

Week1/homework/createTable.js
```diff
@@ -0,0 +1,23 @@
...
+    console.log("Connected!");
+   let tables = [
+        "CREATE TABLE country (Name varchar(50) ,Continent enum('Asia','Europe','North America','Africa','Oceania','Antarctica','South America') ,Region varchar(50),SurfaceArea float, IndepYear int, Population int,  LifeExpectancy float, GNP float ,GNPOld float, LocalName varchar(50), GovernmentForm varchar(50),HeadOfState varchar(50), Capital varchar(50))",
+        "CREATE TABLE city (ID int auto_increment primary key, Name varchar(50), CountryCode varchar(50), District varchar(50), Population int)"
+    ];
+    for (let i in tables) {
+    connection.query(tables[i], function (err, result) {
+        if (err) throw err;
+        console.log("Table created");
+    });
```
mogaleaf: Don't forget to close the connection after the queries otherwise the program keeps running :)

Week1/homework/insertData.js
```diff
@@ -0,0 +1,64 @@
...
+  password : 'hyfpassword',
+  database : 'myWorld'
+});
+
+connection.connect(function(err) {
+    if (err) throw err;
+    console.log("Connected!");
+
+let country = [
+    "insert into country values ('Aruba', 'North America', 'Caribbean', '193.00', NULL, '103000', '78.4', '828.00', '793.00', 'Aruba', 'Nonmetropolitan Territory of The Netherlands', 'Beatrix', 'Oranjestad')",
```
mogaleaf: Yes it's good !

Week1/homework/insertData.js
```diff
@@ -0,0 +1,64 @@
...
+    connection.query(country[i], function(error, results) {
+      if (error) {
+        throw error;
+      }
+      console.log(results[0]);
+    });
+  }
+
+  let city = [
+    "insert into city values(NULL,'Amsterdam', 'NLD', 'Noord-Holland', '731200')",
```
mogaleaf: So I learned something today :) I didn't know you could use NULL to use the Auto increment on mysql !
So It's fine , it works and it's even on the mysql documentation.
BUT personally I don't like it as I find it really confusing and the chances are big that it won't work on an other db like postgres.
What you can do to use the auto increment is giving the name of the column you are filling in the insert and just not provide the ID value , you can read mysql documentation : https://dev.mysql.com/doc/mysql-tutorial-excerpt/5.7/en/example-auto-increment.html

Week1/homework/queries.js
```diff
@@ -0,0 +1,114 @@
...
+});
+
+connection.connect(function(err) {
+    if (err) throw err;
+    console.log('db connected');
+  });
+
+// Question 4i: What are the names of the countries with population greater than 8 million
+let result4i = 'SELECT Name FROM Country WHERE Population > 8000000;'
+
```
mogaleaf: I know you have just copy past but I still don't like the ';' , usually we always avoid unnecessary character.

mogaleaf:
> Dear Cecile, I implemented your feedback as required, I already know that there is no need to use quotes when inserting values into tables but since I am lazy I just copied rows from actual world database in order to populate my own database.
> Additionally, although we weren't taught PK primary key, I settled a PK key in city table because in the homework requirements we have to define an ID for city table, I concluded that if it is an id it must be unique whether I am wrong or not, I don't know.
> Again thank you for your guidance.

Yes Primary Key are always unique, you will see it next week.

mehmet-bozdemir:
Dear Cecile, thanks for your valuable feedback.

mehmet-bozdemir:
For the auto increment method you learned :) I saw it in stackoverflow but your way is better. My solution no doubt can confuse other users.

PR #345, opened 2019-09-10
Mustafa/Week1

mayqf:


mayqf:
a

Week1/index.js
```diff
@@ -0,0 +1,245 @@
+'use-strict';
```
mogaleaf: So I should no read this file right ?
you can use git rm Week1/index.js then commit and push to remove it .

Week1/homework_w1.js
```diff
@@ -0,0 +1,79 @@
...
+
+con.connect(function(err) {
+  if (err) throw err;
+  console.log('Connected to world database!');
+});
+
+var create_queries = [
+  'CREATE DATABASE world',
+  'USE world',
+  'CREATE TABLE country (Name VARCHAR (40) NOT NULL UNIQUE,Continent ENUM("Asia", "Europe", "North America", "Africa", "Oceania", "Antarctica", "South America") DEFAULT "Asia",Region VARCHAR(40) NULL, SurfaceArea INT, IndepYear SMALLINT, Population INT (40) NOT NULL DEFAULT "0",LifeExpectancy FLOAT(4, 1),GNP FLOAT(10,2),GNPOld FLOAT(10,2), LocalName VARCHAR (40),GovernmentForm VARCHAR (30) NULL, HeadOfState VARCHAR(30),Capital VARCHAR(30),PRIMARY KEY (Name))',
```
mogaleaf: Yes it's correct even if Primary Key has not been taught yet.

Week1/homework_w1.js
```diff
@@ -0,0 +1,79 @@
...
+  'CREATE TABLE city (ID INT NOT NULL AUTO_INCREMENT,Name VARCHAR (40) NOT NULL,CountryCode VARCHAR (5),District  VARCHAR(50),  Population INT (40),PRIMARY KEY (ID))',
+];
+
+var insert_queries = [
+  "INSERT INTO country VALUES ('United States', 'North America', 'North America',   9363520.00, 1976,  278357000, 77.1,  8510700.00,  8110900.00, 'United States', 'Federal Republic', 'George W. Bush', 'Washington D.C.')",
```
mogaleaf: Yes it's correct

Week1/homework_w1.js
```diff
@@ -0,0 +1,79 @@
...
+  "INSERT INTO country VALUES ('Argentina', 'South America', 'South America', 2780400.00, 1816, 37032000, 75.1, 340238.00, 323310.00, 'Argentina', 'Federal Republic', 'Fernando de la Ria','Buenos Aires')",
+  "INSERT INTO country VALUES ('Nigeria', 'Africa', 'Western Africa',  923768.00, 1960,  111506000, 51.6,  65707.00 ,  58623.00 , 'Nigeria', 'Federal Republic', ' Olusegun Obasanjo','Abuja')",
+  "INSERT INTO country VALUES ('Georgia', 'Asia', 'Middle East',   69700.00, 1991, 4968000 , 64.5,  6064.00 ,  5924.00 , 'Skartvelo', 'Republic', 'Eduard Âlevardnadze','Tiflis')",
+  "INSERT INTO country VALUES ('Hong Kong', 'Asia', 'Eastern Asia',  1075.00, null , 6782000 , 79.5,  166448.00 , 173610.00 , 'Xianggang/Hong', 'Special Administrative', 'Jiang Zemin','Hong Kong')",
+  "INSERT INTO country VALUES ('Peru', 'South America', 'South America', 1285216.00, 1821 ,  25662000 , 70.0, 64140.00 , 65186.00 , 'Piruw', 'Republic', 'Valentin Paniagua Corazao','Lima')",
+  "INSERT INTO city VALUES (1,'Kabul','AFG','Kabol',1780000)",
```
mogaleaf: It's correct too but as you have used an auto increment for the ID can you find a way to use it here ?

Week1/homework_w1.js
```diff
@@ -0,0 +1,79 @@
...
+  "INSERT INTO city VALUES (10,'Victoria de las Tunas','CUB','Las Tunas',132350)",
+  "INSERT INTO city VALUES (11,'Manzanillo','CUB','Granma',109350)",
+  "INSERT INTO city VALUES (12,'Rotterdam', 'NLD', 'Zuid-Holland',  593321)",
+];
+
+var select_queries = [
+  'SELECT Name, Population FROM country WHERE Population > 8000000',
+  "SELECT Name, Population FROM country WHERE Name LIKE '%land%'",
+  'SELECT Name, Population FROM city WHERE  Population BETWEEN 500000 AND 1000000',
+  "SELECT Name,Continent FROM country WHERE Continent='Europe'",
```
mogaleaf: Why is Continent in column name ? What will be the output ? Is it useful :) ?

Week1/homework_w1.js
mayqf: when I delete id numbers from 'Select...' part it makes error. If I delete auto increment, I can solve the problem. But with auto increment, i could not solve.

Week1/homework_w1.js
mayqf: I know that I filtered country in the Europe. But i wanted to show with continent name.

Week1/homework_w1.js
mogaleaf: If you want to use the auto increment then you have to specify the columns names you want to put the data in otherwise mysql it can't guess which value you put in which column. You don't have to specify column names if you add ALL the value. And with an auto increment you have to remove the ID value to insert, that is why you have this error :)
You can read this for more information on insert : https://www.w3schools.com/sql/sql_insert.asp

Week1/homework_w1.js
mogaleaf: Ok I understand but the value will always be the same Europe :)

mayqf: Yes :)

PR #348, opened 2019-09-10
homework-db-w1-Shyar

ShyarK:


Week1/countryData.js
```diff
@@ -0,0 +1,37 @@
+const checkAndCreate = 'CREATE TABLE IF NOT EXISTS';
+const insert = `INSERT INTO country 
+                (Code, Name, Continent, Region, SurfaceArea, 
+                IndepYear, Population, LifeExpectancy, GNP, GNPOld, 
+                LocalName, GovernmentForm, HeadOfState, Capital) VALUES`;
+const CREATE_COUNTRY_TABLE = `
+                ${checkAndCreate} country(
```
mogaleaf: I love that you are trying to have constant for duplicates.
It would have been more relevant if you could have used this variable for countryData and cityData and not duplicate it.
If you want you can try to have this checkAndCreate const in a common file so you could use it in both query.

Week1/countryData.js
```diff
@@ -0,0 +1,37 @@
...
+                IndepYear  SMALLINT(6) DEFAULT NULL,
+                Population  INT(11) NOT NULL DEFAULT 0,
+                LifeExpectancy  FLOAT(3,1) DEFAULT NULL,
+                GNP  FLOAT(10,2) DEFAULT NULL,
+                GNPOld  FLOAT(10,2) DEFAULT NULL,
+                LocalName  VARCHAR(45) NOT NULL DEFAULT '',
+                GovernmentForm  VARCHAR(45) NOT NULL DEFAULT '',
+                HeadOfState  VARCHAR(60),
+                Capital  VARCHAR(20) NOT NULL,
+                Code2  VARCHAR(2) NOT NULL DEFAULT '',
```
mogaleaf: Code2 :) ?

Week1/countryData.js
```diff
@@ -0,0 +1,37 @@
...
+                ${checkAndCreate} country(
+                Code  VARCHAR(3) NOT NULL DEFAULT '',
+                Name  VARCHAR(52) NOT NULL DEFAULT '',
+                Continent  enum('Asia','Europe','North America','Africa','Oceania','Antarctica','South America') NOT NULL,
+                Region  VARCHAR(26) NOT NULL DEFAULT '',
+                SurfaceArea  FLOAT(10,2) NOT NULL DEFAULT 0.00,
+                IndepYear  SMALLINT(6) DEFAULT NULL,
+                Population  INT(11) NOT NULL DEFAULT 0,
+                LifeExpectancy  FLOAT(3,1) DEFAULT NULL,
+                GNP  FLOAT(10,2) DEFAULT NULL,
```
mogaleaf: It's good but when a column can be NULL you don't have to specify it.

Week1/countryData.js
```diff
@@ -0,0 +1,37 @@
...
+                GNPOld  FLOAT(10,2) DEFAULT NULL,
+                LocalName  VARCHAR(45) NOT NULL DEFAULT '',
+                GovernmentForm  VARCHAR(45) NOT NULL DEFAULT '',
+                HeadOfState  VARCHAR(60),
+                Capital  VARCHAR(20) NOT NULL,
+                Code2  VARCHAR(2) NOT NULL DEFAULT '',
+                PRIMARY KEY (Code)
+              )`;
+
+const INSERT_COUNTRY_QUERIES = `
```
mogaleaf: This is really nice, you have managed to use only one call to the server to insert every lines!

Week1/dbUtils.js
```diff
@@ -0,0 +1,24 @@
+const selectCountry = 'SELECT Name FROM country';
+const selectCity = 'SELECT Name FROM city';
+const SELECT_QUERIES = [
+  `${selectCountry} WHERE Population > 8000000`,
```
mogaleaf: Here it's relevant, you avoid duplicating the String, good !
Your queries are all good.

Week1/countryData.js
ShyarK: Thanks a lot for having time to review my homework! Really appreciate it! Yes, that's right, I'm going to move it to dbUtils file, then I'll require from that file!

PR #350, opened 2019-09-11
Submission Homework Week1

Mehmet-Ergorgec:


Week1/database.js
```diff
@@ -0,0 +1,93 @@
...
+function insertValues(createdQuery) {
+  for (let i = 0; i < createdQuery.length; i++) {
+    createQuery(createdQuery[i]);
+  }
+}
+
+connection.connect();
+
+const create_country =
+  "CREATE TABLE country (Name VARCHAR(60) NOT NULL ,Continent enum('Asia','Europe','North America','Africa','Australia','Antarctica','South America') NOT NULL ,Region VARCHAR(30),SurfaceArea INT NOT NULL, IndepYear INT,Population DECIMAL(5,2) NOT NULL , LifeExpectancy DECIMAL(5,2),GNP BIGINT,GNPOld INT,LocalName VARCHAR(30),GovernmentForm VARCHAR(50),HeadOfState VARCHAR(50),Capital VARCHAR(50))";
```
mogaleaf: Good !

Week1/database.js
```diff
@@ -0,0 +1,93 @@
...
+createQuery(create_country);
+
+const create_city =
+  'CREATE TABLE city (ID INT NOT NULL AUTO_INCREMENT,Name VARCHAR(60) NOT NULL, CountryCode INT NOT NULL, District VARCHAR(60), Population INT NOT NULL , PRIMARY KEY (ID) )';
```
mogaleaf: Good also !

Week1/database.js
```diff
@@ -0,0 +1,93 @@
...
+createQuery(create_city);
+
+const insert_country_queries = [
+  "INSERT INTO country VALUES ('Netherlands', 'Europe', 'Western Europe', 42508 , 1581, 17.08, 81.51, 901900000, NULL, 'Holland', 'Constitutional Monarchy', 'King Willem-Alexander', 'Amsterdam')",
```
mogaleaf: Good also !

Week1/database.js
```diff
@@ -0,0 +1,93 @@
...
+  "INSERT INTO country VALUES ('Turkey', 'Asia', 'Europe', 783562 ,1923, 79.81 ,75.76, 2225000000, NULL,'Türkiye','Republic', 'Recep Tayyip Erdogan','Ankara')",
+  "INSERT INTO country VALUES ('Afghanistan', 'Asia', 'Southern and Central Asia', 652237, 1919, 35.53, 63.67, 71050000, NULL, 'Islamic Emirate of Afghanistan', 'Islamic republic','Ashraf Ghani', 'Kabul')",
+  "INSERT INTO country VALUES ('Canada', 'North America', 'North America', 9985000, 1867, 37.06,82.30, 1679000000,NULL, 'Canada', 'Constitutional Monarchy', 'Queen Elizabeth II','Ottowa')",
+  "INSERT INTO country VALUES ('Germany', 'Europe', 'Europe',357386, 1990, 82.79, 80.64, 4280000000 , NULL, 'Deutschland', 'Republic','Frank-Walter Steinmeier', 'Berlin')",
+];
+insertValues(insert_country_queries);
+
+const insert_city_queries = [
+  "INSERT INTO city SET name='Amsterdam', CountryCode=31, District='North Holland', Population= 821752",
```
mogaleaf: I like that you tried something different, yes it's working and you can reuse this way to run update queries too.
Plus you are using the Auto increment that way!

Week1/database.js
```diff
@@ -0,0 +1,93 @@
...
+
+const continentsInEurope =
+  "SELECT name AS continents_in_Europe FROM country WHERE Continent='Europe'";
+createQuery(continentsInEurope);
+
+const countrySurfaceAreaDesc =
+  'SELECT name,SurfaceArea AS countries_surfaceArea_in_desc FROM country ORDER BY SurfaceArea DESC';
+createQuery(countrySurfaceAreaDesc);
+
+const citiesInNetherlands =
```
mogaleaf: It's nice you tried the Join !
It this case it's not really useful because you still have no "connection" between city table and country table. It will come next week with the foreign key.
Instead of the JOIN you could also have select on both table city and country done a WHERE clause with condition on both table. Want to try ?

Mehmet-Ergorgec: I tried JOIN because I could not find another solution. I did not find any similarities between city and country tables especially to get cities in Netherlands in order to match and use condition in query. Can you give me a hint:)

Mehmet-Ergorgec:
Thank you @mogaleaf for your kind review;
I completely forget to send query about creating and using database because of I was working on created database in mysql server. But I implemented and commited those changes.
Best regards;

Week1/database.js
mogaleaf: The Join is absolutely fine don't worry, it's just an other way of doing it that I find maybe more readable for this query but an other developer could have an other opinion.

You can select on multiple table and use condition to retrieve the correct rows .
Example with two tables mytable1 and mytable2:
SELECT mytable1.columA,mytable2.columnB FROM mytable1,mytable2 WHERE mytable1.columnC = 'something' AND mytable2.columnD = 'something else';
Actually it's even called implicit join (I had to google it :) )

So here you can simply select from city and country and do a where clause where city.CountryCode=31 and country='netherlands' :)

PR #351, opened 2019-09-11
Adham data base homework week1

Adham-Elias-Botours:
here is my first homework "database week1":
I think that I did not understand the files and folders structure (tree) in the repository. Also it was not clear where I have to write my code.
So I made  a new folder "homework" , it contains my work.

The content of "homework" folder:

- create_db.js: you have to start with this file to create the database and it's data. Executing this file might takes about 10 minutes, because it is going to add more than 4000 records to the data.

- info_from_data.js: This file executes MYSQL queries to select specific information from the data. So you can run it after the first file.

- queries_db_tb.js: this file contains MYSQL queries in javascript objects. You don't have to run this file, but it is required as a resource to build the data.
Thanks


mogaleaf:
Hi Adham,

Thanks for the explanation, it's always useful for the reviewer !

I guess all of this data come from world.sql. They are almost good for the exercice but use notion that have not been taught yet (Primary Key, Foreign Key and Auto increment) .

I propose you to try to review the Create statement to only column that have been asked so it's a bit easier (no foreign key for instance except if you know the concept then I'm ok with it). 
Pay an extra attention to some datatype that are not exactly the one asked ( Capital for instance is a int in world.sql but I think it should be a varchar ? )

And then insert to let say 15 countries and 5 cities ? 

Depending of the changes you may have to rewrite some of the select queries.
Tell me if everything is clear :)

Adham-Elias-Botours:
Dear @mogaleaf 
I made the requested changes.  I think all of them ...
In my work: the table country has a field "code". That was not requested in the homework question, but it is the only way to make a kind of relation between this table and the city table.

Thanks

Week1/homework/queries_db_tb.js
```diff
@@ -0,0 +1,81 @@
+'use strict';
+
+const dropDbWorld = `DROP SCHEMA IF EXISTS world`;
+const creatDbWorld = `CREATE SCHEMA world`;
+
+const dropTableCountry = `DROP TABLE IF EXISTS country`;
+const CreateTableCountry = `CREATE TABLE  country  (
```
mogaleaf: Good !

Week1/homework/queries_db_tb.js
```diff
@@ -0,0 +1,81 @@
...
+  ID INT(11) NOT NULL AUTO_INCREMENT,
+  Name VARCHAR(35) NOT NULL DEFAULT '',
+  CountryCode CHAR(3) NOT NULL DEFAULT '',
+  District VARCHAR(20) NOT NULL DEFAULT '',
+  Population INT(11) NOT NULL DEFAULT '0',
+  PRIMARY KEY (ID)
+)`;
+
+const insertContires = [
+    "INSERT INTO country VALUES ('USA','United States','North America','North America',9363520.00,1776,278357000,77.1,8510700.00,8110900.00,'United States','Federal Republic','George W. Bush','WASHINGTON, D.C.')",
```
mogaleaf: Good also !

Week1/homework/queries_db_tb.js
```diff
@@ -0,0 +1,81 @@
...
+    "INSERT INTO country VALUES ('CHN','China','Asia','Eastern Asia',9572900.00,-1523,1277558000,71.4,982268.00,917719.00,'Zhongquo','Peoples Republic','Jiang Zemin','BEIJING')",
+    "INSERT INTO country VALUES ('BRA','Brazil','South America','South America',8547403.00,1822,170115000,62.9,776739.00,804108.00,'Brasil','Federal Republic','Fernando Henrique Cardoso','BRASILIA')",
+    "INSERT INTO country VALUES ('IND','India','Asia','Southern and Central Asia',3287263.00,1947,1013662000,62.5,447114.00,430572.00,'Bharat/India','Federal Republic','Kocheril Raman Narayanan','NEW DELHI')",
+    "INSERT INTO country VALUES ('ARG','Argentina','South America','South America',2780400.00,1816,37032000,75.1,340238.00,323310.00,'Argentina','Federal Republic','Fernando de la Rúa','BUENOS AIRES')",
+    "INSERT INTO country VALUES ('KAZ','Kazakstan','Asia','Southern and Central Asia',2724900.00,1991,16223000,63.2,24375.00,23383.00,'Qazaqstan','Republic','Nursultan Nazarbajev','NUR-SULTAN')",
+];
+
+const insertCities = [
+    "INSERT INTO city VALUES (ID,'Amsterdam','NLD','Noord-Holland',731200)",
```
mogaleaf: Yes it works and you even used the Auto Increment.
An other solution would have been to use the columns names where you insert the data :
INSERT table_name (col1,col2,..) VALUES (c1,c2,...). This way you can select which values you want to insert and you can use also Auto Increment or default Values.

Week1/homework/queries_db_tb.js
mogaleaf: If you want to go a bit further you could try to do only ONE INSERT query but entering multiple rows. It helps reducing the number of calls on the DB.

Week1/homework/info_from_data.js
```diff
@@ -0,0 +1,41 @@
...
+    if (err) console.log(err);
+    else console.log(`Connected!to database: world`);
+});
+const infoFromData = [
+    'SELECT name FROM country WHERE population > 8000000 ',
+    'SELECT name FROM country WHERE name LIKE "%land%"',
+    'SELECT name FROM city WHERE population BETWEEN 500000 AND 1000000',
+    'SELECT name FROM country WHERE Continent = "Europe"',
+    'SELECT name FROM country ORDER BY SurfaceArea DESC;',
+    'SELECT city.name AS NederlandsCities FROM city INNER JOIN country ON city.CountryCode = country.Code WHERE country.name = "Netherlands"',
```
mogaleaf: It's good but INNER keyword is not necessary, JOIN is INNER by default .

Week1/homework/queries_db_tb.js
Adham-Elias-Botours: I am going to read about that.. interesting.

Week1/homework/info_from_data.js
Adham-Elias-Botours: Screenshot_20190911-215807_1
Is this explanation in the photo correct?  then all the words in ( ) are not necessarily, Like what you mentioned.

Adham-Elias-Botours: thanks,
I post a link here as a reference for me. 
https://www.w3resource.com/mysql/advance-query-in-mysql/mysql-joins.php

Week1/homework/queries_db_tb.js
Adham-Elias-Botours: I leave also a reference for me, 
https://www.w3schools.com/nodejs/nodejs_mysql_insert.asp

PR #352, opened 2019-09-12
submit week1 homework solutions - Salih S

salih18:
In the solutions I insert values from JSON files which include all country and city information. 

I tried to implement by using basic connection method however there are so many data has to been inserted, i changed it to pool connection method to reduce waiting time. Before, all operation has completed at least 5 minutes, but after using pool connection it takes seconds to complete all queries.

The homework includes bonus parts, also I added one more question to try.

The index.js file fires up all operations.


Week1/homework/databaseClasses.js
```diff
@@ -0,0 +1,118 @@
...
+      LifeExpectancy FLOAT,
+      GNP FLOAT,
+      GNPOld FLOAT,
+      LocalName VARCHAR(60) ,
+      GovernmentForm VARCHAR(60),
+      HeadOfState VARCHAR(80),
+      Capital varchar(80),
+      Code2 VARCHAR(2)
+    ) DEFAULT CHARSET = utf8mb4;`;
+
```
mogaleaf: I guess you did put this encoding on purpose. I didn't know about it but apparently you are right, it is recommended to encode correctly.

mogaleaf: Your create statement is really good.
I know that in this first week lesson you have seen teacherS table but it's more common to use singular for table name, like here city and country. But it's just a convention.

Week1/homework/databaseClasses.js
```diff
@@ -0,0 +1,118 @@
...
+          country.Population,
+          country.LifeExpectancy,
+          country.GNP,
+          country.GNPOld,
+          country.LocalName,
+          country.GovernmentForm,
+          country.HeadOfState,
+          country.Capital,
+          country.Code2,
+        ]);
```
mogaleaf: This is really good

Week1/homework/databaseClasses.js
```diff
@@ -0,0 +1,118 @@
...
+  async createCitiesTable() {
+    try {
+      const q = `CREATE TABLE IF NOT EXISTS cities (
+        ID INT AUTO_INCREMENT PRIMARY KEY,
+        Name VARCHAR(60),
+        CountryCode VARCHAR(60),
+        District VARCHAR(60),
+        Population INT   
+      ) DEFAULT CHARSET = utf8mb4;`;
+
```
mogaleaf: Good also

Week1/homework/databaseClasses.js
```diff
@@ -0,0 +1,118 @@
...
+      console.log('Cities table could not create ' + e.stack);
+    }
+  }
+
+  async insertCities() {
+    try {
+      cities.forEach(async city => {
+        const q = 'INSERT INTO cities SET Name=?, CountryCode=?, District=?, Population=?';
+        await pool.query(q, [city.Name, city.CountryCode, city.District, city.Population]);
+      });
```
mogaleaf: yes good too :)

Week1/homework/query.js
```diff
@@ -0,0 +1,51 @@
...
+const q9 = "9. What's the top 10 cities with the highest population?";
+const qu9 = 'SELECT Name, Population FROM cities ORDER BY Population DESC LIMIT 10';
+const q10 = "10. What's the population of the world ?";
+const qu10 = 'SELECT SUM(Population) AS World_Population FROM countries';
+const q11 = '11. What are the largest populated countries in each continents';
+const qu11 =
+  'SELECT Name, Population, Continent FROM countries WHERE Population IN (SELECT MAX(Population) FROM countries WHERE Population > 1 GROUP BY Continent)';
+
+const queries = [qu1, qu2, qu3, qu4, qu5, qu6, qu7, qu8, qu9, qu10, qu11];
+const questions = [q1, q2, q3, q4, q5, q6, q7, q8, q9, q10, q11];
```
mogaleaf: Instead of dealing with two arrays, maybe a dictionary could be easier here ? A key/value structure ?

Week1/homework/query.js
```diff
@@ -0,0 +1,51 @@
...
+const q8 = "8. What's the top 10 countries based on surface area?";
+const qu8 = 'SELECT Name, SurfaceArea FROM countries ORDER BY SurfaceArea DESC LIMIT 10';
+const q9 = "9. What's the top 10 cities with the highest population?";
+const qu9 = 'SELECT Name, Population FROM cities ORDER BY Population DESC LIMIT 10';
+const q10 = "10. What's the population of the world ?";
+const qu10 = 'SELECT SUM(Population) AS World_Population FROM countries';
+const q11 = '11. What are the largest populated countries in each continents';
+const qu11 =
+  'SELECT Name, Population, Continent FROM countries WHERE Population IN (SELECT MAX(Population) FROM countries WHERE Population > 1 GROUP BY Continent)';
+
```
mogaleaf: It's really good! I tried it myself and came with the same kind of solution. You could also have used UNION to do it if you want to give a try.

Week1/homework/index.js
```diff
@@ -0,0 +1,29 @@
...
+      await city.createCitiesTable();
+      await city.insertCities();
+
+      queries.forEach((query, index) => {
+        result.answer(query, questions[index]);
+      });
+    } catch (e) {
+      console.log(e.stack);
+    }
+  }
```
mogaleaf: The connection / pool is not closed at the end. I'm not familiar with node js but if you could find a way to release the connection at the end ?

salih18: As you know, the advantage of using pool connection instead of using single connection is that we can run queries in parallel and it enables us to run multiple queries in a short time. Normally, when you run a NodeJS web application, you never think about shutting down your app because it should run forever but in our case after run multiple queries we have to close the database connection. The side effect of the using pool connection method of mysql module is that it doesnt have graceful end() feature to shutdown the connection. I spent 2 days just implement why it is like that, so many people are complaining about that and they suggest graceful end to overcome this problem by using mysql module.

When you use single connection and run queries then connection.end() method waits the return results of the queued queries and it closes the connection, whereas pool.end() method doesnt wait for the queries result, it just close the connection and you will get an error ''Pool is closed.." Yes there must be a way to get rid of this problem in JS world, but in MYSQL package there is not a built in way.

Week1/homework/query.js
salih18: Thank you.. Firstly I have to learn what UNION is :) I will look about it..

Week1/homework/query.js
salih18: Yes, that would be better. I have just spent my days to implement closing the pool connection in a proper way.

Week1/homework/databaseClasses.js
salih18: If i didnt put charset it would throw errors in some columns.

Week1/homework/index.js
mogaleaf: What about something like :
```suggestion
// create a list of promises
     const promises = queries.map((query, index) =>  result.answer(query, questions[index]));
//wait them all to fulfill
      await Promise.all(promises);
// close the pool
       pool.end();

  }
```
in index.js ?

salih18: Wow. Thank you for the your crystal clear solution. It works perfectly. :smile:

mogaleaf: I read this to find the solution :
https://medium.com/@patarkf/synchronize-your-asynchronous-code-using-javascripts-async-await-5f3fa5b1366d

PR #357, opened 2019-09-14
finished homework for databases week01

Laurelinduilwen:
might have been a bit sloppy with this one, realized I havent fixed the "var"s from the template too late. might have also messed up the naming conventions will fix asap...  

no terminals were hurt during the creation of this homework...

### DIFF
--- a/Week1/Week01.js
+++ b/Week1/Week01.js
@@ -1,0 +1,48 @@
+var mysql = require('mysql');
+var connection = mysql.createConnection({
+  host: 'localhost',
+  user: 'hyfuser',
+  password: 'hyfpassword',
+});
+
+connection.connect();
+const database = 'create database if not exists world';
+const useDatabase = 'use world';
+
+const countries =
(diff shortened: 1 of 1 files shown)


Week1/Week01.js
```diff
@@ -0,0 +1,48 @@
...
+  password: 'hyfpassword',
+});
+
+connection.connect();
+const database = 'create database if not exists world';
+const useDatabase = 'use world';
+
+const countries =
+  'create table country (name varchar(50), continent varchar(50), region varchar(50), surface_area int, independence_year date, population int, gnp int, gnp_old int, local_name varchar(50), government_form varchar(50), head_of_state text, capital varchar(50))';
+const cities =
```
mogaleaf: Be careful some values can not be int as they require floating point :) Could you fix them ? 
Otherwise your naming is perfect!

Laurelinduilwen: fixed them :)

Laurelinduilwen:
Thank you for the feedback :) Sorry for being late with the implementation,,,

PR #359, opened 2019-09-17
Database week 1

elicarcar:


Week1/homework/queries.js
```diff
@@ -0,0 +1,26 @@
...
+  host: 'localhost',
+  user: 'hyfuser',
+  password: 'hyfpassword',
+  database: 'myWorld',
+});
+
+const queries = [
+  `select name from country where population > 8000000;`,
+  `select name from country where name like '%land%';`,
+  `select name from city where population > 50000 or population < 1000000;`,
```
mogaleaf: Be careful here you get population > 5000 OR population < 1000000. What if my population is 5000000 ? 
It will be selected it because 5000000 > 5000 but you don't want to select this row. Can you fix this condition?

Week1/homework/create-tables.js
```diff
@@ -0,0 +1,29 @@
...
+  host: 'localhost',
+  user: 'hyfuser',
+  password: 'hyfpassword',
+  database: 'myWorld',
+});
+
+connection.connect();
+
+const create_country_query =
+  "create table country (Name VARCHAR(20), Continent enum('Asia', 'Europe', 'North America', 'Africa', 'Oceania', 'Antarctica', 'South America'), Region text, Surface_Area float, IndepYear date, Population INT, Life_Expectancy INT(3), GNP float, GNPOld float, Local_Name text, Government VARCHAR(25), Head_Of_State VARCHAR(60), Capital text)";
```
mogaleaf: Good !

Week1/homework/create-tables.js
```diff
@@ -0,0 +1,29 @@
...
+connection.query(create_country_query, function(error, results, fields) {
+  if (error) {
+    throw error;
+  }
+  console.log('your table is' + results);
+});
+
+const create_city_query =
+  'create table city (name varchar(20), country_code char(3), district text, population int)';
+connection.query(create_city_query, function(error, results, fields) {
```
mogaleaf: Don't forget the ID :) But Its good otherwise

Week1/homework/insert-values.js
```diff
@@ -0,0 +1,51 @@
+const mysql = require('mysql');
+const conn = mysql.createConnection({
+  host: 'localhost',
+  user: 'hyfuser',
+  password: 'hyfpassword',
+  database: 'myWorld',
+});
+
+conn.connect();
+const insert_country_queries = [
```
mogaleaf: Inserts are good

Week1/homework/queries.js
elicarcar: So true, im fixing it

Week1/homework/create-tables.js
elicarcar: ok i added the id as well, but is there another way to automatize the id without typing it all the time?

mogaleaf: Yes you can use auto increment so it will be incremented by default. But you need to add which column name you fill in your insert to use it :)

PR #360, opened 2019-09-17
Week2/adil

musafir928:
Add homework step 1 cli version.


Week2/homework/index.js
```diff
@@ -0,0 +1,170 @@
...
+from city 
+where id 
+in (select capital from country where name = ?);`
+
+const searchRegion = `select distinct language 
+from countrylanguage 
+where countrycode 
+in(select code from country where region = ?);`
+
+const searchNumOfCities = `select distinct name
```
mogaleaf: I understand what you have done giving the array length. But what if there is a lot of data, is it worth it to consume all this memory just to give the length of an array ? Maybe you can find a better solution using sql ?

Week2/homework/index.js
```diff
@@ -0,0 +1,170 @@
...
+from city 
+where countrycode 
+in (select countrycode from countrylanguage where language = ?);`
+
+const searchOfficialLanguage = `select name 
+from country 
+where region = ? 
+and code in(select countrycode from countrylanguage where language = ? and isofficial = 't');`;
+
+const getContinentList = `select distinct continent
```
mogaleaf: Ok It works using too queries, but you can do it in once :) Think about the group by statement.

musafir928:
> Thanks Adil for you homework.
> You have great ideas and I like it ! Let's try to use a bit more sql to help us doing less in the server :)
> I don't see part 2 and part 3 but I guess you'll add it later?

Thank you for your reviewing and feedback. Yes, I will push them soon.

Week2/homework/index.js
musafir928: fixed! As you said I found a very simple way, thank you.

Week2/homework/index.js
musafir928: Ok, I will make it simple.

musafir928: Thanks, during changing my code I have learned a lot.

PR #363, opened 2019-09-19
Adham data base homework week 2

Adham-Elias-Botours:
Here is my homework for Database week 2.
Before this : Commits on Sep 18, 2019, I have been working by mistake on the master branch.

I have nothing more to say, I am exhausted !

But I tomorrow I have to prepare few questions for the next lesson.
Thanks


Week2/homework/index.js
```diff
@@ -0,0 +1,117 @@
...
+                }
+            case options.fourth:
+                const userInput4 = await inquirer.prompt([
+                    { name: 'regionName', message: 'What is the region?' },
+                    { name: 'languageName', message: 'What is the language?' },
+                ]);
+                const query4 = `SELECT Name FROM country join countrylanguage on country.Code=countrylanguage.CountryCode where Region =? and Language=? and IsOfficial='T'; `;
+                let result4 = await queryPromise(query4, [userInput4.regionName, userInput4.languageName]);
+                if (result4.length === 0) {
+                    console.log(noMatch);
```
mogaleaf: Here it should be false but it's fine that way ( false because the user could enter two good input but still no countries speak this language).

Week2/homework/index.js
```diff
@@ -0,0 +1,117 @@
...
+
+const noMatch = 'NO match is found in the database!\nBe sure that you typed a valid word';
+
+connection.connect(function(err) {
+    if (err) console.log(err);
+});
+
+(async() => {
+    try {
+        const userChoice = await inquirer.prompt([{
```
mogaleaf: Your menu works perfectly.

Week2/homework/index.js
```diff
@@ -0,0 +1,117 @@
...
+            message: 'make your choice?',
+            choices: [options.first, options.second, options.third, options.fourth, options.fifth],
+        }, ]);
+
+        switch (userChoice.mainChoice) {
+            case options.first:
+                const userInput1 = await inquirer.prompt([
+                    { name: 'countryName', message: 'What is the country?' },
+                ]);
+                const query1 = `SELECT city.Name FROM country JOIN city on city.ID = country.Capital WHERE country.Name = ?`;
```
mogaleaf: good query

Week2/homework/index.js
```diff
@@ -0,0 +1,117 @@
...
+                } else {
+                    console.log(`The capital is: ${result1[0].Name}`);
+                    break;
+                }
+
+            case options.second:
+                const userInput2 = await inquirer.prompt([
+                    { name: 'regionName', message: 'What is the region?' },
+                ]);
+                const query2 = `SELECT distinct Language FROM country join countrylanguage on country.Code=countrylanguage.CountryCode where Region =?`;
```
mogaleaf: good query too

Week2/homework/index.js
```diff
@@ -0,0 +1,117 @@
...
+                    console.log(`The languages are:`);
+                    result2.forEach(element => console.log(element.Language));
+                    break;
+                }
+
+            case options.third:
+                const userInput3 = await inquirer.prompt([
+                    { name: 'languageName', message: 'What is the language?' },
+                ]);
+                const query3 = `SELECT  count(city.name) as Cities FROM city join countrylanguage on city.CountryCode=countrylanguage.CountryCode where Language=? `;
```
mogaleaf: also this one is good

Week2/homework/index.js
```diff
@@ -0,0 +1,117 @@
...
+                } else {
+                    console.log(`The number of cities is: ${result3[0].Cities}`);
+                    break;
+                }
+            case options.fourth:
+                const userInput4 = await inquirer.prompt([
+                    { name: 'regionName', message: 'What is the region?' },
+                    { name: 'languageName', message: 'What is the language?' },
+                ]);
+                const query4 = `SELECT Name FROM country join countrylanguage on country.Code=countrylanguage.CountryCode where Region =? and Language=? and IsOfficial='T'; `;
```
mogaleaf: and this query also good

Week2/homework/index.js
```diff
@@ -0,0 +1,117 @@
...
+                    console.log(noMatch);
+                    break;
+                } else {
+                    console.log(`The counties are:`);
+                    result4.forEach(element => console.log(element.Name));
+                    break;
+                }
+
+            case options.fifth:
+                const query5 = `SELECT Continent, COUNT(DISTINCT Language) as Languages FROM country join countrylanguage on country.Code=countrylanguage.CountryCode GROUP BY Continent`;
```
mogaleaf: yes !

Week2/homework/trigger_countrylanguages.sql
```diff
@@ -0,0 +1,11 @@
+CREATE DEFINER=`hyfuser`@`localhost` TRIGGER `countrylanguage_AFTER_INSERT` AFTER INSERT ON `countrylanguage` FOR EACH ROW BEGIN
```
mogaleaf: your trigger is correct, but it would have been better to define it BEFORE insert and not AFTER so mysql has less work to do

Adham-Elias-Botours:
> Hi Adham,
> 
> Queries are perfect nice job!
> Your trigger is good two I would have just use it BEFORE the insert than AFTER.
> 
> The only thing is on the ER diagram, I don't understand why you have list_01 and list_02 ? Two tables for the lists ? Furthermore if I look the fields it seems to be more tasks than list ? Could you explain me a bit more what you have tried to do ?
> 
> Thanks a lot

[The ERD question.pdf](https://github.com/HackYourFuture/databases/files/3637594/The.ERD.question.pdf)
Dear @mogaleaf ,
In the pdf file you will find a long explanation about what I have done with the ERD answer. 
I know it is wrong, and I agree with you about that. But I explained my way in order  to get a clear and direct feedback form you.
Also could you please rewrite the question to make it easier for me,  and then write your own analyse. I need to understand how I have to think and analyze such a thing.

Then I have to upgrade my work (fix it) in few days , as soon as possible. 

thanks.

Week2/homework/index.js
Adham-Elias-Botours: Thanks for accepting that. 
If I am the user , then the word "False"  is meaningless to  mee. I might think there is something wrong in the program.  maybe it is not meaningless for the one who requested it.

Week2/homework/trigger_countrylanguages.sql
Adham-Elias-Botours: yes now I got the point.
On the other hand: the way how the question is written makes the idee not clear (for me). 
The question was: If a country X has 9 languages in the CountryLanguage table, and a user INSERTs one more row in the CountryLanguage table, then I should get an alert.

mogaleaf:
Thanks for the PDF it helped me understand.

You usually decide the model( meaning which tables will be created so your system can work) and it's usually static. By this I mean you create the model and all your tables before even writing the code. 
Then you write the code that INSERT data in this tables. 

So the biggest problem with your approach is that you want to create table each time a user wants to add a list. It's a lot of work and not really easy to maintain. How could you get all the tables for one user ? You can query rows but tables name it's hard... 

But you are almost there, why instead of creating a new table each time a user wants a list, you would not just add a row in list table ? referring the user it belongs to ?
And same for taks, Task can be a table and each row refer the list it belongs to .

No need trigger here, do something simple, try to respond to this questions :
 - When I want a list in which table my INSERT should go, how can I find my list after the insert ?
 - When I want a task in which table my INSERT should go , how can I find my task after the insert ?

I hope I have been helpful !

Adham-Elias-Botours:
> * When I want a list in which table my INSERT should go,

Thanks a lot,
Now I have a very clear idea about how I should answer that question. 
In short: I do not have to sort and save the data. But I have to save the data in a structure that makes another process simple and easy,  I mean process like adding, deleting, updating, sorting or selecting  info.
So here: 2 main tables are enough (lists table, tasks table). Then 2 another tables (task reminders table, task tags table).
And I think my next ERD going to be good. I will upload it as soon as possible.

Adham-Elias-Botours:
Dear @mogaleaf 
My work is adjusted. Now with new ERD and the trigger is adjusted (trigger before insert, not after).
I think I am on the right way.
thanks

mogaleaf:
Perfect now ! I validate your homework !

Adham-Elias-Botours:
> Perfect now ! I validate your homework !

Thanks a lot.

PR #366, opened 2019-09-20
Add Homework week 2

gulyener:
This week's homework was very challenging for me. I could only finish part 1 and part 2, I will work on part 3 and hopefully finish it before sunday.


Week2/hw-part1.js
```diff
@@ -0,0 +1,137 @@
...
+  let inputLanguage = '';
+  let inputRegOf = '';
+  let inputLangOf = '';
+
+  try {
+    const onSubmit = (prompt, response) => console.log(`You typed ${colors.rainbow(response)}`);
+    const response = await prompts(questions, { onSubmit });
+
+    inputCountryCode = response.CountryCode.toUpperCase();
+    inputRegion = response.region
```
mogaleaf: Did you have problem with the Case in your DB ? usually by default it's case insensitive so you don't have to pay attention to upper case at the beginning of input word. But otherwise the code is good

Week2/hw-part1.js
```diff
@@ -0,0 +1,137 @@
...
+      .toLowerCase()
+      .split(' ')
+      .map(s => s.charAt(0).toUpperCase() + s.substring(1))
+      .join(' ');
+    inputLangOf = response.language2.replace(/^\w/, c => c.toUpperCase());
+
+    const selectCapital =      'SELECT name FROM city WHERE ID =(SELECT capital FROM country WHERE Code = ?);';
+    const selectLang =      'SELECT Language FROM countrylanguage JOIN country ON CountryCode = Code WHERE region = ?';
+    const selectNum = 'SELECT count(countryCode) FROM countrylanguage WHERE Language = ?';
+    const selectOfficial = `SELECT countryCode, name AS Country, isOfficial FROM country JOIN countryLanguage ON countryCode = Code WHERE region = '${inputRegOf}' AND language = '${inputLangOf}'`;
```
mogaleaf: I think you misunderstood the question in the homework. You print False if you don't find any official language. You have printed False if the language is not official. So just add the isOfficial condition in your where statement and it's good :)

Week2/hw-part1.js
```diff
@@ -0,0 +1,137 @@
...
+
+    // Query 3
+    console.log(
+      `Q3. Number of cities ${inputLanguage} is spoken: ${colors.magenta(
+        results3[0]['count(countryCode)'],
+      )}`.green,
+    );
+
+    // Query 4
+    console.table(
```
mogaleaf: See my comment on the query selectOfficial

Week2/hw-part1.js
```diff
@@ -0,0 +1,137 @@
...
+      if (results4[i].isOfficial == 'T') {
+        console.log(`Official in: ${results4[i].Country}`);
+      } else {
+        console.log('FALSE');
+      }
+    }
+
+    // Query 5
+    console.log('Q5. Continents and number of languages spoken:'.green);
+    console.table(results5);
```
mogaleaf: You forgot to print the continent and numbers here :)

Week2/hw-part1.js
gulyener: No I only did it for my eye's pleasure:D But you are right it's unnecessary I removed it :)

Week2/hw-part1.js
gulyener: Thanks! I actually understood the question right but I couldn't make it work, so I did the next best thing:) But now it works and I am happy thank you

Week2/hw-part1.js
gulyener: I actually wanted to display it as a table. I guess that was not the intention of the homework. Now I changed it.

PR #369, opened 2019-09-21
Submission Homework Week2

Mehmet-Ergorgec:
Part 3 is now in progress. I am going to do and commit it asap.


Week2/database.js
```diff
@@ -0,0 +1,112 @@
...
+const getLanguagesInRegion = async () => {
+  try {
+    const response = await prompts({
+      type: 'text',
+      name: 'region',
+      message: `List all the languages spoken in the region ...?`,
+    });
+    const userInput = response.region.charAt(0).toUpperCase() + response.region.slice(1);
+    const regionQuery = await execQuery(question2, userInput);
+    const validateQuery = await execQuery(`select region from country group by region`);
```
mogaleaf: Why this query ? You could just check the length of regionQuery to see if you have language .

Week2/database.js
```diff
@@ -0,0 +1,112 @@
...
+};
+
+const getLanguagesInRegion = async () => {
+  try {
+    const response = await prompts({
+      type: 'text',
+      name: 'region',
+      message: `List all the languages spoken in the region ...?`,
+    });
+    const userInput = response.region.charAt(0).toUpperCase() + response.region.slice(1);
```
mogaleaf: You don't need this, sql is not case sensitive.

Week2/database.js
```diff
@@ -0,0 +1,112 @@
...
+const getTotalCityWithLanguage = async () => {
+  try {
+    const response = await prompts({
+      type: 'text',
+      name: 'language',
+      message: `Find out how many cities speak this language?`,
+    });
+    const userInput = response.language.charAt(0).toUpperCase() + response.language.slice(1);
+    const totalCitiesSpeakLanguage = await execQuery(question3, response.language);
+    const validateQuery = await execQuery(`select language from countrylanguage`);
```
mogaleaf: Same here no need, just check the length of the response.

Week2/database.js
```diff
@@ -0,0 +1,112 @@
...
+const getOfficialOrNot = async () => {
+  const response = await prompts({
+    type: 'list',
+    name: 'value',
+    message: 'Are there any countries uses this language officially or not?',
+    initial: 'Ex:Middle East,Turkish',
+    separator: ',',
+  });
+  const isOfficial = await execQuery(question4, response.value);
+  const trueOfficial = isOfficial.filter(item => item.officialOrNot === 'T');
```
mogaleaf: this can be done inside the sql query to avoid work here.

Week2/queries.js
```diff
@@ -0,0 +1,68 @@
+const question1 = `
+SELECT
```
mogaleaf: good

Week2/queries.js
```diff
@@ -0,0 +1,68 @@
...
+const question2 = `
+   SELECT
+      countrylanguage.language AS language 
+   FROM
+      country,
+      countrylanguage 
+   WHERE
+      countrylanguage.countrycode = country.code 
+      AND country.region = ? 
+   GROUP BY
```
mogaleaf: No need of group by here

Week2/queries.js
```diff
@@ -0,0 +1,68 @@
...
+      language 
+   ORDER BY
+      language`;
+
+const question3 = `
+      SELECT
```
mogaleaf: good

Week2/queries.js
```diff
@@ -0,0 +1,68 @@
...
+        AND country.code = countrylanguage.countrycode 
+        AND countrylanguage.language = ? 
+      GROUP BY
+        language`;
+
+const question4 = `SELECT
+        region,
+        language,
+        country.name AS country,
+        isofficial as officialOrNot
```
mogaleaf: Add the condition in the WHERE clause

Week2/queries.js
```diff
@@ -0,0 +1,68 @@
...
+        countrylanguage
+     WHERE
+        countrylanguage.countrycode = country.code
+        AND region = ?
+        AND language = ? `;
+
+const question5 = `
+        SELECT
+           continent,
+           count(language) AS total_spoken_language
```
mogaleaf: What if english is spoken in canada and USA and you choose America ? It will be counted twice

Week2/trigger.js
```diff
@@ -0,0 +1,19 @@
+DELIMITER $$
+  CREATE TRIGGER max_language
```
mogaleaf: good

Week2/database.js
Mehmet-Ergorgec: I thought something different. But you are right.Thank you

Week2/database.js
Mehmet-Ergorgec: When I write this code which is Js , Mysql did not accept lower case. So I implemented this code. But I have changed it. Now no need. Thank you

Week2/queries.js
Mehmet-Ergorgec: There are some duplicated regions so I group them.

Mehmet-Ergorgec:
Thank you for your  kind review. I implemented all your feedbacks. But in ERD Diagram each tag,reminder or mark should refer listItem instead of list. In homework I understand like that. I am looking forward to hear feedback about ERD because I will use it Homework week 3 after change some logic as well:)

Week2/queries.js
mogaleaf: Ok got it, you could have used a distinct also on countrylanguage .

Mehmet-Ergorgec:
I have not inserted any listItem. I just created tables.because of that you did not see any listItem.  I understand that if I dont have any listItem in my list there is no need for reminder and tag. After I insert listItem in my list it is more meaningful to have reminder and tag.
If it is so, should I create my reminder,tag and mark tables without fk? should I insert fk after inserted any listItem on my list?

mogaleaf:
Hoo I think I understand now what you mean by listItem !! It's rows in List table ! so called them List :) In user table we insert user so each row is a user. In List table you insert List so each row is a list.
Now Each list should contains multiple item (that was my misunderstanding with listitem). So you need a Item table and this Item table needs a foreign key on list table. One item is attached to One list.  One list can have multiple items. 
Now each item can have one reminder and one tag to make it simple.

So you need : 
List table primary key id => You have it , foreign key on User table and you have it
Item table primary key id => You need to add it. Item table has a foreign key on List table. you can call it list_id
Reminder table primary key id => You have it. Reminder table has a foreign key on Item Table, you can call it item_id
Tags table primary key id => You have it. Tags table has a foreign key on Task table, you can call it item_id.

Is it better ?

Mehmet-Ergorgec:
I am sorry @mogaleaf now I notice your message because of homework week3. I think there is a conflicts in naming. :) I understand what you mean. Thank you

PR #372, opened 2019-09-24
Submit Week3 Homework - Oguzhan

oguzhanturgut:
- **TODO_API.md** Installation and usage of API
- **helpers.js** Helper functions for middlewares. All SQL queries are built top on Knex module.
- **validate.js** Post body validation using [Joi](https://hapi.dev/family/joi/?v=16.0.1) schemas
- **routes** folder Endpoint routing for `/user` `/list` and `/tag`
- **todo.sql** Mysql dump file of todo schema. Don't have too much generic dummy data. Sorry about that.

For standalone [repository](https://github.com/oguzhanturgut/todo-api) and [README](https://github.com/oguzhanturgut/todo-api/blob/master/README.md)

Further information about [Knex](http://knexjs.org/)

I know some parts seem like spaghetti code. Tried my best for refactoring but couldn't complete because of time problem.


Week3/TODO_API.md
```diff
@@ -0,0 +1,53 @@
...
+| ---------------------------------------- | ------------------------------------ | ---------------------------------------------------------------- | --------------------------------------- |
+| `GET /list/:listid`                      | Get a list by list id                |                                                                  |                                         |
+| `DELETE /list/:listid`                   | Delete a list by listid              |                                                                  |                                         |
+| `GET /list/:listid/todo`                 | Get all Todos of a list by listid    |                                                                  |                                         |
+| `GET /list/:listid/todo/:todoid`         | Get Todo by todoid                   |                                                                  |                                         |
+| `POST /list/:listid/todo`                | Add new Todo to a list by list id    | `{`<br/>`"note": "Buy milk",`<br/>`"tags": "food,daily"`<br/>`}` | add multiple tags using comma(optional) |
```
mogaleaf: tag with no S otherwise it's not working :)

Week3/TODO_API.md
```diff
@@ -0,0 +1,53 @@
...
+| `DELETE /list/:listid/todo/:todoid`      | Delete a Todo by todo id             |                                                                  |                                         |
+| `POST /list/:listid/todo/:todoid/done`   | Mark Todo as done by todo id         |                                                                  |                                         |
```
mogaleaf: Here I would not have use a delete for the done but an other endpoint as done is not an entity
post /list/:listid/todo/:todoid/done = done
post /list/:listid/todo/:todoid/undo = not done

Week3/TODO_API.md
oguzhanturgut: Fixed typo. Thanks

Week3/TODO_API.md
```diff
@@ -0,0 +1,53 @@
...
+
+`npm start`
+
+**! Do not forget to change database login credentials in `helpers.js` !**
+
+## Usage
+
+### *`/user` endpoint*
+
+| **Request**               | **Description**                 | **Body**                                                                            | **Details**                  |
```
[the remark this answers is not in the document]
oguzhanturgut: Thank you

Week3/TODO_API.md
oguzhanturgut: Changed the HTTP verb to PATCH since making a partial modify following the [convention](https://www.restapitutorial.com/lessons/httpmethods.html).
`PATCH /list/:listid/todo/:todoid/done` done
`PATCH /list/:listid/todo/:todoid/undo`  not done

Thanks for the tip

oguzhanturgut:
> Hello,
> 
> Good jobs, and nice this knex library !
> I added some comments on the endpoints. I think developer could speak hours on what is the best structure to have about endpoint, but try to use only one way so the user is not confused when using your api.
> All your error code are almost perfect !
> 
> About the architecture of the code, you have separated the router in three sections and it makes sense ! So why not doing the same with the handlers ?
> A helper should only contains really basic method to help the project but should not contains logic, for instance a helper could have method to generate random number, or convert dates to string, or parse the tags. But avoid having the entire app in a helper because if the app is getting bigger then no one will understand where to look to maintain it.
> 
> So I would suggest to either spread the different handlers with the files containing the route, or add new files with list_handler.js, users_handler.js and so one.
> 
> But it's a good job I validate the homework.

Hello @mogaleaf 

I seperated the `helpers.js` and spread handlers into `handlers` directory with endpoint names. `helpers.js` has only helper functions not whole logic now. Also all handlers and `helpers.js` are getting `knex` constructor from `db.js`.
I really appreciate you for making time and effort to give quite precious comments on the homework. Thank you 👏☺️

PR #373, opened 2019-09-24
Database-homeworkw3 --Shyar

ShyarK:


Week3/api.js
```diff
@@ -0,0 +1,218 @@
...
+      res.send(user);
+    } catch (err) {
+      console.log(err);
+    }
+  }
+
+  async createUserHandler(req, res) {
+    try {
+      const { Name, User_id, Email } = req.body;
+      const create = `INSERT INTO todos_app.users (User_id, Name, Email) VALUES (?, ?, ?)`;
```
mogaleaf: User_id could have been auto increment, it's hard to ask people to provide their own primary key.

Week3/README/READEME.md
```diff
@@ -0,0 +1,56 @@
...
+       "item_id": 2,
+       "description": "going to hyf amsterdam",
+       "todo_completed": "true",
+       "todolist_id": 2
+  }``
+   
+ <h3>The users object:</h3>
+ 
+ ``{
+        "user_id": 7,
```
mogaleaf: It's a capital U here otherwise it's not working

Week3/README/READEME.md
```diff
@@ -0,0 +1,56 @@
...
+
+`http://localhost:3000/todoitems`
+
+<h3> To get one user with the given id:</h3>
+
+`http://localhost:3000/user/7`
+
+<h3> To get one todoList with the given id:</h3>
+
+`http://localhost:3000/todolist/2`
```
mogaleaf: Usually you present the doc like:
http://localhost:3000/todolist/{id}, then you can add an example.

Week3/api.js
```diff
@@ -0,0 +1,218 @@
...
+connection.connect(handelError('connecting...'));
+
+class Api {
+  async getUsersHandler(req, res) {
+    try {
+      const users = await execQuery(`SELECT * FROM todos_app.users`);
+      if (!users.length) return res.status(404).send(`You need to insert some users!`);
+      res.send(users);
+    } catch (err) {
+      console.log(err.message);
```
mogaleaf: If you get an error you have to return an error code (500) so the client knows something wrong happened.

Week3/api.js
```diff
@@ -0,0 +1,218 @@
...
+    }
+  }
+
+  async deleteUserHandler(req, res) {
+    try {
+      const users = await execQuery(`SELECT * FROM todos_app.users`);
+      const id = parseInt(req.params.id);
+      const user = users.find(user => user.user_id === id);
+      if (!user) return res.status(404).send('The user with the given Id was not Found!');
+      if (users.length === 1) await execQuery(`Delete FROM todos_app.users where user_id = ?`, id);
```
mogaleaf: Ok but if you want you can have a look to DELETE ON CASCADE feature in mysql, when you delete user, it will delete all underlying row that have a foreign key on user.

Week3/api.js
```diff
@@ -0,0 +1,218 @@
...
+      res.end('The update was successful!');
+    } catch (err) {
+      res.end(err);
+    }
+  }
+
+  async getTodolistHandler(req, res) {
+    try {
+      const todolist = await execQuery(`SELECT * FROM todos_app.todolist`);
+      if (!todolist.length) return res.status(404).send(`Create a Todo List!`);
```
mogaleaf: I think it's ok not to have todolist yet ? I would respond 404 if you try to get a specific id but not if you want a list.

Week3/api.js
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  async getSpecificTodolistHandler(req, res) {
+    const todolistId = req.params.id;
+    try {
+      const result = await execQuery(`SELECT * FROM todos_app.todolist WHERE todolist_id = ?`, [
+        todolistId,
+      ]);
+      if (!result.length)
+        return res.status(404).send('The Todo List with the given Id was not Found!');
```
mogaleaf: Yes here it's a correct 404

Week3/app.js
```diff
@@ -0,0 +1,35 @@
...
+const app = express();
+const bodyParser = require('body-parser');
+const Api = require('./api');
+const api = new Api();
+
+// parse application/json
+app.use(bodyParser.json());
+
+// users
+app.get('/users', api.getUsersHandler);
```
mogaleaf: your routes are good !

Week3/api.js
ShyarK: Thank you very much to review my homework and give feedback! 
I don't understand but I have auto increment: 

``CREATE TABLE `users` (
  `user_id` int(11) NOT NULL AUTO_INCREMENT,
  `Name` varchar(45) DEFAULT NULL,
  `Email` varchar(60) DEFAULT NULL,
  PRIMARY KEY (`user_id`)
)``
I don't understand here what to change? Can you please give more detail?

Week3/README/READEME.md
ShyarK: I don't know if I get you right! You mean I need to change it in MySQL or in README.md file?
because I have it like thins:
```user_id` int(11) NOT NULL AUTO_INCREMENT``

mogaleaf: This is not sql here it's read the json keys.
The json keys are case sensitive and when you read the key later you read with User_id.
 > const { Name, User_id, Email } = req.body;
So either you change the readme or the code but it has to be consistent :)

Week3/api.js
mogaleaf: Yes you have it in the db but you don't use it, if you want to use it, either you put null at User_id or you provide the column name in your insert and you don't give id column. then sql will fill the field by itself.

Week3/README/READEME.md
ShyarK: I didn't know that we need to use ``{id}``, in the doc! As I know so far we use template literals or string literals to allow embedded expressions: 
For example: 
``const person = 'Mike';``
``const age = 28;``
``const output = myTag`That ${ person } is a ${ age }`;``

And also we use it ``react.js`` to embed expressions in JSX:
``const name = 'Josh Perez';``
``const element = <h1>Hello, {name}</h1>;``
You asked to use it in the doc file! Can you give an example so I can change it? Actually, this is my first time to write README.md file.

Week3/api.js
ShyarK: OK, I got the point, 
 const { Name, Email } = req.body;
 const create = `INSERT INTO todos_app.users (Name, Email) VALUES (?, ?)`;
 await execQuery(create, [Name, Email]);

Week3/README/READEME.md
mogaleaf: You don't NEED to use but it's conventional. I'm not sure there is one standard but you can use parameters in {name} or :name .
Here is one of github api documentation:
https://developer.github.com/v3/orgs/members/

Usually documentation are generated from code by annotated the code itself.

The documentation won't affect your code, it's just a way to explain the user how to perform a call. Then you can provide example of call to help the developer understand.

PR #375, opened 2019-09-27
Mustafa/Week3

mayqf:
- ToDo App database(multiple users)

- Node server to make Web APIs for ToDo app

- Support following functions:

1. Insert item(s) in ToDo list
2. Delete item(s) in ToDo list
3. Create a new ToDo list
4. Delete a ToDo list
5. Mark an item as completed
6. Add a reminder for the list (not for the item)
---------------------------------------------------------
connection.js=>used createPool() for making connection to db.
user_queries.js=> defined for todo user queries 
list_queries.js=> defined for todo list queries 
ist_item.js=> defined for todo item queries
server.js=>used express() for server


Week3/homework3/item_queries.js
```diff
@@ -0,0 +1,73 @@
...
+    res.send('Todo item deleted successfully');
+  } catch (err) {
+    console.log(err);
+  }
+};
+const markAsFinished = async (req, res) => {
+  try {
+    const id = parseInt(req.params.id);
+    const todocompletions = await execQuery(`SELECT * FROM todo_app.list_item_completion`);
+    const completion = todocompletions.find(completion => completion.ID === id);
```
roxana-lafuente: Can you please double check that this one works good? I've added a todo, then I tried to mark it as done but although it is getting the right ID, it always says ID not found.

PR #378, opened 2019-09-27
Db week 3

elicarcar:
Couldn't load the dump file due to complicated reasons but I will try to add it asap.


Week3/db-week-3.js
```diff
@@ -0,0 +1,216 @@
...
+      throw err;
+    }
+  }
+});
+
+app.post('/saveAddItem', async (req, res) => {
+  try {
+    const connection = mysql.createConnection(todoDatabase);
+    const execQuery = util.promisify(connection.query.bind(connection));
+    const itemQuery = `INSERT INTO items VALUES ('${req.body.task}', '${req.body.isdone}', '${req.body.listName}')`;
```
roxana-lafuente: Just one more option is to use it like this:

```
const task = ${req.body.task};
const isDone = ${req.body.isdone};
const listName = ${req.body.listName};

const itemQuery = `INSERT INTO items VALUES (?, ?, ?)`;
await execQuery(itemQuery, [task, isDone, listName]);
```

Simply because this way:

- sentences are shorter
- more readable for other developers

Week3/instructions/Readme.md
```diff
@@ -0,0 +1,30 @@
...
+- Type the path and then for deleting an item: /deleteItem/<--itemname-->
+- For deleting a list /deleteList/<--listname-->
+- You will get the response that says 'Succesfully deleted'
+
+FOR UPDATING THE STATUS OF AN ITEM
+
+- Open you postman
+- Chose your method as PUT
+- Type the path and then for deleting an item: /update/<itemYouWantToUpdate>/<updatedStatus>
+- You will get the response that says 'Your item is up to date'
```
roxana-lafuente: Let's imagine you're developing this app for a company, they will need to use the APIs you developed but they don't know where to start. They read your documentation and they know where to start, they need postman. But they don't know all the functionalities you provide, then they need search the source code you wrote. But there's a problem, they don't know javascript, node or anything similar. And they don't know how to use your amazing API.

That's why it's a good idea to document all the functionality available, which parameters each one receive. Here is an example of a [good API documentation](https://docs.gitlab.com/ee/api/api_resources.html#project-resources) 

If you add a table like this, your documentation would be perfect! :)

Week3/package.json
```diff
@@ -0,0 +1,21 @@
+{
```
roxana-lafuente: Great idea to add the package.json, this makes installation of your code dependencies super straightforward.

Week3/db-week-3.js
```diff
@@ -0,0 +1,216 @@
...
+      throw err;
+    }
+  }
+});
+
+app.put('/update/:item/:isdone', async (req, res) => {
+  try {
+    const connection = mysql.createConnection(todoDatabase);
+    const execQuery = util.promisify(connection.query.bind(connection));
+    const updateQuery = `UPDATE items SET isdone = '${req.params.item}' Where items.item = '${req.params.isdone}';`;
```
roxana-lafuente: When we develop APIs, we always need to return proper HTTP Codes so that the person / program using our API can understand what is happening. I mean something like this:

```
if (!${req.params.item})
       return res.status(404).send('ID not Found!');
```

Here's more info on the HTTP Codes https://www.restapitutorial.com/httpstatuscodes.html

Could you please add them?

Week3/db-week-3.js
```diff
@@ -0,0 +1,216 @@
...
+  password: 'hyfpassword',
+  database: 'todo',
+  multipleStatements: true,
+};
+
+app.set('view engine', 'pug');
+app.use(bodyParser.urlencoded({ extended: false }));
+app.use(express.static(__dirname + '/public'));
+
+const CREATE_TABLE_TODOS = `CREATE TABLE IF NOT EXISTS todos (
```
roxana-lafuente: Please don't forget to add the sql dump file

Week3/db-week-3.js
```diff
@@ -0,0 +1,216 @@
...
+    belongs_to VARCHAR(30),
+    
+    FOREIGN KEY (belongs_to)
+    REFERENCES todos(todoid)
+    ON DELETE CASCADE
+    );`;
+
+app.get('/', async function(req, res) {
+  try {
+    const connection = mysql.createConnection(todoDatabase);
```
roxana-lafuente: No need to change it, but there's also this option https://stackoverflow.com/questions/37102364/how-do-i-create-a-mysql-connection-pool-while-working-with-nodejs-and-express to use it across your code. Might be worth looking into it for future projects :)

Week3/db-week-3.js
```diff
@@ -0,0 +1,216 @@
...
+    }
+  }
+});
+
+app.get('/download', function(req, res) {
+  const file = `${__dirname}/instructions/Readme.md`;
+  res.download(file);
+});
+
+app.post('/submit', async function(req, res) {
```
roxana-lafuente: It's usually a good practice to have self descriptive names, for instance:

submitReview, submitForm, submitHomework, etc instead of just submit.

The idea is that the person reading the code has an idea of what your function does without the need to read the body of the function itself. The same with variable names :)

Week3/db-week-3.js
```diff
@@ -0,0 +1,216 @@
...
+      throw err;
+    }
+  }
+});
+
+app.delete('/deleteList/:list', async (req, res) => {
+  try {
+    const connection = mysql.createConnection(todoDatabase);
+    connection.query(
+      `DELETE FROM todos where todos.todoid = ?`,
```
roxana-lafuente: Just be careful since here you're hardcoding the database name `database_name.table_name` but if you change it in the configuration above, it will not work anymore :)

elicarcar:
@roxana-lafuente  thank you for your feedback! And so sorry for the late implementation and response, i wll try to finish it asap!

Week3/db-week-3.js
```diff
@@ -0,0 +1,297 @@
...
+    }
+  }
+});
+
+
+
+app.put('/update/:item/:isdone', async (req, res) => {
+  try {
+    const item = req.params.item;
+    const isDone = req.params.isDone;
```
roxana-lafuente: Here it should be:

```
const isDone = req.params.isdone;
```

(lowercase d)

Week3/db-week-3.js
```diff
@@ -0,0 +1,297 @@
...
+    const connection = mysql.createConnection(todoDatabase);
+    const execQuery = util.promisify(connection.query.bind(connection));
+    const updateQuery = `UPDATE items SET isdone = ? Where items.item = ?;`;
+    execQuery(updateQuery, [item, isDone]).then(res.status(201).send('Your item is up-to-date!'));
```
roxana-lafuente: And here it should be:

```
execQuery(updateQuery, [isDone, item]).then(res.status(201).send('Your item is up-to-date!'));
```

PR #381, opened 2019-09-28
mehmet bozdemir database wek3 (new)

mehmet-bozdemir:


Week3/homework/app.js
```diff
@@ -0,0 +1,150 @@
...
+// Create todolist
+const createTodoList = (req, res) => {
+  const user_id = req.body.user;
+  const todolist_name = req.body.name;
+  const createTodolist = `INSERT INTO todolist (user_id, todolist_Name) VALUES (?, ?)`;
+  connection.query(createTodolist, [user_id, todolist_name], err => {
+    if (err) {
+      throw err;
+    }
+    res.status(200).send('Todolist created!');
```
roxana-lafuente: Perfect that you're always sending the proper HTTP codes! This is super important :)

Week3/homework/app.js
```diff
@@ -0,0 +1,150 @@
...
+    }
+    return res.status(201).send('Todo item created!');
+  });
+};
+
+// Detete todos item
+const deleteTodoItem = (req, res) => {
+  const todo_id = req.params.id;
+  if (!todo_id) {
+    return res.status(400).send('Todo not found!');
```
roxana-lafuente: Great error handling. Just one detail, you want 404 here I believe since the error message is "todo not found"

Week3/homework/help.md
```diff
@@ -0,0 +1,65 @@
+Database Homework Week3 Readme
```
roxana-lafuente: Lovely documentation. You can also use markdown language, more info here: https://en.support.wordpress.com/markdown-quick-reference/

Week3/homework/app.js
```diff
@@ -0,0 +1,150 @@
...
+  database: 'todo',
+});
+
+// Create User
+const createUser = (req, res) => {
+  const user_name = req.body.name;
+  if (!user_name) {
+    return res.status(400).send('Please enter a username for user!');
+  }
+  const createUser = `INSERT INTO Users (username) VALUES (?);`;
```
roxana-lafuente: Careful here: Creating the same user twice breaks the server

Week3/homework/app.js
```diff
@@ -0,0 +1,150 @@
...
+    if (err) {
+      return res.status(500).send('Something went wrong');
+    }
+  });
+  return res.status(201).send('New user successfully created!');
+};
+
+//Get users
+const getUsers = (req, res) => {
+  const getUsersQuery = `SELECT * from users`;
```
roxana-lafuente: If this was a production database, we should try to avoid doing SELECTs like this without limits because if we had billions of users, we don't generally want to do this. It might break the page, or it can be very slow!

Week3/homework/app.js
```diff
@@ -0,0 +1,150 @@
...
+    }
+    res.json(results);
+  });
+};
+
+// Update User
+const updateUser = (req, res) => {
+  const user_id = req.params.id;
+  const user_name = req.body.name;
+  const updateUsersQuery = `UPDATE users SET username = ?  WHERE user_id = ${user_id}`;
```
roxana-lafuente: username is not a valid field in this table :) did you mean Name?

Week3/homework/app.js
```diff
@@ -0,0 +1,150 @@
...
+    }
+    res.status(200).send('Todo item marked as done');
+  });
+};
+
+app.post('/create/user', createUser);
+app.get('/users', getUsers);
+app.get('/users/:id', getUser);
+app.put('/users/:id', updateUser);
+app.post('/create/todolist', createTodoList);
```
roxana-lafuente: Adding a get for users was a great idea because the user of the API does not have access to the DB unless it is through this API. It would have been nice to add it for todo lists and items too, do you think you can add it? :)

Week3/homework/app.js
```diff
@@ -0,0 +1,150 @@
...
+app.delete('/todolist/:id', deleteTodoList);
+app.post('/create/todo', createTodo);
+app.delete('/todo/:id', deleteTodoItem);
+app.put('/todo/:id', markAsDone);
```
roxana-lafuente: For future projects, try to split the code into different modules instead of a single file :)

PR #386, opened 2019-10-11
Week3/adil

musafir928:


Week3/homework/server.js
```diff
@@ -0,0 +1,190 @@
...
+    con.query(
+      `insert into users values (?,?,?)`,
+      [result.length + 1, req.body.username, req.body.password],
+      (err, data) => {
+        if (err) {
+          res.render('error', { data: 'error' });
+        } else {
+          res.render('index');
+        }
+        con.end();
```
mogaleaf: If you close the connection then any queries run after this code won't work and will generate connection closed issue.
This comment is for all queries

Week3/homework/server.js
```diff
@@ -0,0 +1,190 @@
...
+        },
+      );
+    },
+  );
+});
+
+//  show lists
+app.get('/user/lists', urlParser, (req, res) => {
+  con.query(
+    `select * from items inner join list on items.listid = list.id where uid in (select id from users where username='${req.query.username}') group by name;`,
```
mogaleaf: this query is not working, you can't group by name , and you don't need too, username should be uniq in your case otherwise that means you bring list for an other user.
By the way, do you want to bring list or items ? because if I don't have item then I don't bring anything..

Avoid using ${req.query.username} directly in the query but instead use prepared statement ? otherwise your code becomes vulnerable to attack.

Week3/homework/server.js
```diff
@@ -0,0 +1,190 @@
...
+      con.end();
+    },
+  );
+});
+
+// delete list
+app.get('/user/lists/delete/', urlParser, (req, res) => {
+  const listName = req.query[Object.keys(req.query)];
+  const queries = [
+    `delete from list where name='${listName}';`,
```
mogaleaf: same here the code is highly vulnerable to attack, I could drop the database

PR #387, opened 2019-10-28
Msonmez hw week1

msnmz:


Week1/index.js
```diff
@@ -0,0 +1,173 @@
...
+    fields.length > 1 ? fields.join(',') : fields[0]
+  } from ${table} ${filter}`;
+  try {
+    const result = await asyncQuery(query);
+    console.log('\n\nResult of ', filter, '\n');
+    console.table(result);
+  } catch (err) {
+    showErrorAndExit(`Filtering Error on (${filter}): ${err.message}`, connection);
+  }
+}
```
mogaleaf: I love you tried to avoid repetition in your code, but what do you think is the best: 
 - People can read the code easily and we have some repetition
 - Or you don't have any repetition but to understand the query you have to log it.

Personally I'm the first kind of developer. I always try to avoid repetition but sometimes it's better to keep them for code readability.

Week1/index.js
```diff
@@ -0,0 +1,173 @@
...
+const connection = mysql.createConnection({
+  host: 'localhost',
+  user: 'root',
+  password: 'password',
+});
+
+const asyncQuery = promisify(connection.query.bind(connection));
+
+// Queries in order, to be executed
+const queries = [
```
mogaleaf: it's a great choice to use query message structure.
Your create statements are all good.

Week1/index.js
```diff
@@ -0,0 +1,173 @@
...
+  // 1- names of all the cities in the Netherlands?
+  await queryAndShowResultOf(`where country_code = 'NLD'`, 'city', ['*']);
+  // 2 - population of Rotterdam
+  await queryAndShowResultOf(`where name = 'Rotterdam'`, 'city', ['population']);
+  // 3 - top 10 countries based on surface area
+  await queryAndShowResultOf(`order by surface_area desc limit 10`, 'country', ['*']);
+  // 4 - top 10 cities with the highest population
+  await queryAndShowResultOf(`order by population desc limit 10`, 'city', ['*']);
+  // 5 - population of the world
+  await queryAndShowResultOf('', 'country', ['sum(population) as world_population']);
```
mogaleaf: Your queries are all good!

msnmz: Thank you very much for your kind explanations and review :)
I tried to avoid repetition for the sake of homework but I agree that sometimes avoiding repetitions makes code less readable :(

PR #391, opened 2019-11-01
Db week1

husenjanobul:


Week1/hm/queries4.js
```diff
@@ -0,0 +1,39 @@
...
+});
+
+connection.connect();
+
+const queries =[
+    // countries with population greater than 8 million
+    'SELECT name FROM country WHERE population>=8000000',
+
+    // countries that have “land” in their names
+    'SELECT name FROM country WHERE name LIKE "%lands"',
```
roxana-lafuente: All queries are OK except this one but I think it's just a typo and you meant:

```
SELECT name FROM country WHERE name LIKE "%land%"
```

:)

Week1/hm/queries4.js
```diff
@@ -0,0 +1,39 @@
...
+
+    // cities with population in between 500,000 and 1 million
+    'SELECT name FROM city WHERE population>=500000 and population<= 1000000',
```
roxana-lafuente: Another option would have been to use between. More info here: https://www.techonthenet.com/mysql/between.php

Week1/hm/queries5.js
```diff
@@ -0,0 +1,39 @@
...
+    host: 'localhost',
+    user: 'hyfuser',
+    password : 'hyfpassword',
+    database: 'world'
+});
+
+connection.connect();
+
+const queries =[
+    // cities with population in between 500,000 and 1 million
```
roxana-lafuente: Careful here: the comments don't match what the queries do :)

Week1/hm/createDatabase.js
```diff
@@ -0,0 +1,14 @@
+const mysql = require('mysql');
+const connection = mysql.createConnection({
+    host: 'localhost',
+    user: 'hyfuser',
+    password : 'hyfpassword',
+})
+
+connection.connect();
```
roxana-lafuente: It's always a good idea to add validation on everything that could go on. And connecting to the DB is one of those places where tons of things can go wrong: user does not exist, wrong password, database does not exist, etc

Week1/hm/queries5.js
husenjanobul: Thank you a lot for review and feedback . 
I’ll check again. :)

PR #392, opened 2019-11-01
Homework week1

ahmthsrvsmz:


Week1/homework/index.js
```diff
@@ -0,0 +1,205 @@
...
+});
+
+connection.connect();
+
+const createCountries = `CREATE TABLE countries (
+  Name varchar(254) ,
+  Continent varchar(50),
+  Region varchar(50) ,
+  SurfaceArea float(2) ,
+  IndepYear varchar(50) ,
```
mogaleaf: We could argue here with an int :)

Week1/homework/index.js
```diff
@@ -0,0 +1,205 @@
...
+  Population float(2)  ,
+  LifeExpectancy varchar(50) ,
```
mogaleaf: Same here I think int should be better

Week1/homework/index.js
```diff
@@ -0,0 +1,205 @@
...
+  GNP float(2) ,
+  GNPOld varchar(50) ,
```
mogaleaf: Why gnp float and not gnpOld ?

Week1/homework/index.js
```diff
@@ -0,0 +1,205 @@
...
+        console.log(`Cities name with "land"`);
+        res.send(results);
+      }
+    },
+  );
+});
+
+app.get('/cityPopulation', (req, res) => {
+  connection.query(
+    `SELECT * FROM cities WHERE Population >500000 and Population <1000000;`,
```
mogaleaf: If you want you can also use between here

Week1/homework/index.js
```diff
@@ -0,0 +1,205 @@
...
+      }
+      console.log(`Cities logged`);
+      res.send(results);
+    },
+  );
+});
+
+app.get('/haveLandName', (req, res) => {
+  connection.query(
+    `SELECT * FROM countries WHERE countries.Name like "%land%" ;`,
```
mogaleaf: good

Week1/homework/index.js
```diff
@@ -0,0 +1,205 @@
...
+        console.log(`Cities that have population in between 500,000 and 1 million`);
+        res.send(results);
+      }
+    },
+  );
+});
+
+app.get('/countryContinent', (req, res) => {
+  connection.query(
+    `SELECT * FROM countries WHERE Population >500000 and Population <1000000;`,
```
mogaleaf: I think it's not the good query here, it was supposed to be :
names of all the countries on the continent ‘Europe’

Week1/homework/index.js
```diff
@@ -0,0 +1,205 @@
...
+        console.log(`Cities that have population in between 500,000 and 1 million`);
+        res.send(results);
+      }
+    },
+  );
+});
+
+app.get('/surface_area', (req, res) => {
+  connection.query(
+    `SELECT * FROM countries ORDER BY SurfaceArea DESC;`,
```
mogaleaf: good

Week1/homework/index.js
```diff
@@ -0,0 +1,205 @@
...
+  }
+  res.write(`Table has values`);
+  res.end();
+});
+
+app.post('/appendCities', async (req, res) => {
+  try {
+    const { cities } = data;
+    for (const city of cities) {
+      const try1 = `INSERT INTO cities values(NULL, '${city.name}', '${city.country_code}', '${city.district}', ' ${city.population}'); `;
```
mogaleaf: Good for the use of auto-increment.
After next week you will have to use prepared query to avoid attack on your code :)

Week1/homework/index.js
```diff
@@ -0,0 +1,205 @@
...
+    console.log(`Table created`);
+    res.send(`Table created`);
+  });
+});
+
+app.post('/appendCountries', async (req, res) => {
+  try {
+    const { countries } = data;
+    for (const country of countries) {
+      const try1 = `INSERT INTO countries VALUES('${country.name}', '${country.continent}', '${country.region}', '${country.surface_area}', ' ${country.indep_year}', ' ${country.population}', ' ${country.life_expectancy}', ' ${country.gnp}', ' ${country.gnp_old}', ' ${country.local_name}', ' ${country.government_form}', ' ${country.head_of_state}', ' ${country.capital}');`;
```
mogaleaf: good for this week

Week1/homework/index.js
ahmthsrvsmz: Okay, we will :)

Week1/homework/index.js
ahmthsrvsmz: Mysql gave me an error after that I changed it. 👍  I will try again with float.

Week1/homework/index.js
ahmthsrvsmz: got it 👍

Week1/homework/index.js
ahmthsrvsmz: Thank you :)

Week1/homework/index.js
ahmthsrvsmz: my fault :( missed it.

Week1/homework/index.js
ahmthsrvsmz: Thanks:)

Week1/homework/index.js
ahmthsrvsmz: okay I will :) Thanks 👍

Week1/homework/index.js
ahmthsrvsmz: Thanks :)

Week1/homework/index.js
ahmthsrvsmz: Table hasn't values Error: ER_TRUNCATED_WRONG_VALUE_FOR_FIELD: Incorrect integer value: ' null' for column 'IndepYear' at row 1 (When int)

Week1/homework/index.js
mogaleaf: I think you should not use '' around the value if it's float or integer because it is used for string (varchar). Try with removing the quote when inserting into the db to see if you are still getting error.

ahmthsrvsmz: Yes, you are right. I can use float(2). It is okay now. I will update it again. :)

PR #400, opened 2019-11-09
Ahmet week2

ahmthsrvsmz:


Week2/homework/index.js
```diff
@@ -0,0 +1,125 @@
...
+        console.log(error);
+        res.send(error);
+      } else {
+        res.send(results);
+      }
+    },
+  );
+});
+
+app.get(`/countries/:languages`, (req, res) => {
```
mogaleaf: This way work but only if the user is adding data from your program. The user would still be able to add a new country using the sql command line for instance. That is why mysql provides a trigger. Can you try to write one instead?

Week2/homework/index.js
```diff
@@ -0,0 +1,125 @@
...
+});
+
+connection.connect();
+
+const data = require('./cities.json');
+const { countries } = data;
+const { cities } = data;
+
+function getCapitalCity(id, req, res) {
+  connection.query(`select * from cities where id = "${id}"`, (error, results, fields) => {
```
mogaleaf: Avoid injecting user data like that :"${id}". Use prepared statement instead ( where id = ? ).

ahmthsrvsmz: 👍

Week2/homework/index.js
ahmthsrvsmz: 👍

Tjebbee:
@mogaleaf ; is this one ready as well?

Tjebbee:
Please @ahmthsrvsmz , take a look at this

PR #403, opened 2019-11-09
mehmetekinci-db-week2-hw

mehmetekinci:


Week2/index.js
```diff
@@ -0,0 +1,92 @@
...
+    return;
+  } else if (questionNumber < 1 || questionNumber > 5) {
+    console.error('Please enter a number between 1 and 5');
+    return;
+  }
+  const queries = [
+    {
+      query: `select city.name as CAPITAL from city join country on 
+      city.id=country.capital where country.name='${answer1}';`,
+      number: 1,
```
mogaleaf: The query is good but you have to use Prepared statement please (no country.name='${answer1}' but instead country.name= ? to avoid sql injection

Week2/index.js
```diff
@@ -0,0 +1,92 @@
...
+    },
+    {
+      query: `select language,region from country join countrylanguage on 
+      country.code=countrylanguage.countryCode where region = '${answer1}';`,
+      number: 2,
```
mogaleaf: Good but same as query 1

Week2/index.js
```diff
@@ -0,0 +1,92 @@
...
+    },
+    {
+      query: `select COUNT(city.name) from city join countrylanguage on 
+      city.countryCode=countrylanguage.countryCode where language = '${answer1}';`,
+      number: 3,
```
mogaleaf: Good but same as query 1

Week2/index.js
```diff
@@ -0,0 +1,92 @@
...
+    },
+    {
+      query: `select country.name from country join city on country.capital=city.id
+      join countrylanguage on city.countryCode=countrylanguage.countryCode where
+      region='${answer1}' and language='${answer2}' and isOfficial='T';`,
```
mogaleaf: Good but same as query 1

Week2/index.js
```diff
@@ -0,0 +1,92 @@
...
+      number: 4,
+    },
+    {
+      query: `select continent, COUNT(language) from country join countrylanguage
+       on country.code=countrylanguage.countryCode GROUP BY continent;`,
+      number: 5,
```
mogaleaf: What if a continent as more than one country speaking one language? For instance, North America has the USA and Canada with the English language. You should count English only once :)

Week2/index.js
```diff
@@ -0,0 +1,92 @@
...
+
+const showQuery = async (questionNo, answer1, answer2) => {
+  const questionNumber = parseInt(questionNo, 10);
+  if (Number.isNaN(questionNumber)) {
+    console.error('Please enter a number.');
+    return;
+  } else if (questionNumber < 1 || questionNumber > 5) {
+    console.error('Please enter a number between 1 and 5');
+    return;
+  }
```
mogaleaf: A good check of the user's actions!

Week2/index.js
mehmetekinci: @mogaleaf  Thank you for your helpful feedback. I made it.

Week2/index.js
mehmetekinci: Yes. I see. I found the DISTINCT. it works.

mehmetekinci:
@mogaleaf I also uploaded  part2. Thank you for the great feedbacks and compliments :) .

PR #411, opened 2019-11-16
Db week3 hw

EsraZengn:
Here is my databases week3 homework. Looking forward to your feedbacks.

### DIFF
--- a/Week3/homework/README.md
+++ b/Week3/homework/README.md
@@ -1,0 +1,226 @@
+# HackYourFuture Databases Week 3 Homework: Full ToDo App
+
+This app supports following functions:
+• create user(s)
+• create todoList(s) for user
+• create todoListItem(s) for todoList
+• create reminder(s) for todoList
+• delete user
+• delete todoList of user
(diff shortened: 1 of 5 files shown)


EsraZengn:
Thank you very much Cecile Thiebaut  for the feedback. What you pointed out was actually the things on my mind. I will implement your suggestion about returning elements with id right away, as you said it is important to see the id for later use and it is a good idea to add it to the response. About the second suggestion, actually this is the second version of my homework. At first i tried to divide app into smaller parts kind of like the one we did for the node.js week3 homework, but just when I was about to finish -which was yesterday night- I realized there was something wrong with my way of implementing and the structure was not efficient at all. So I decided to rewrite my hw as a single file since there was a deadline. Can you possibly help me with the other one and maybe tell me the right way to do it? I was gonna ask on slack but maybe you can help me better since you know my homework already. Again many thanks for the feedbacks.

mogaleaf:
Yes, of course, you can push the previous version and I will have look at it. And you can point out what you think is wrong in it.

PR #416, opened 2019-11-17
Ahmet week3

ahmthsrvsmz:


Week3/homework/index.js
```diff
@@ -0,0 +1,128 @@
...
+  password: 'password',
+  database: 'todo_database',
+});
+
+connection.connect();
+
+app.put(`/createList/:list`, (req, res) => {
+  const { list } = req.params;
+  connection.query(
+    `CREATE TABLE ${list} (todo_id INT(11) NOT NULL AUTO_INCREMENT,  todo varchar(50), done enum("true", "false"), user_id INT,  PRIMARY KEY (todo_id)  );`,
```
mogaleaf: What is the purpose of asking the user the name of the table? you should have a table list containing a user_id which is a foreign key on user table, this way you know to which user belongs the list.
You could have a list table like you have with todo and fill it with insert statement.

Week3/homework/index.js
```diff
@@ -0,0 +1,128 @@
...
+        res.send(results);
+      }
+    },
+  );
+});
+
+app.post(`/addUser`, (req, res) => {
+  const { name, email } = req.body;
+  connection.query(
+    `insert into User_List values (null, "${name}", "${email}"); `,
```
mogaleaf: Please use prepared statement, not direct string injection in the request otherwise your are vulnarable to hackers.

Tjebbee:
@ahmthsrvsmz ; please implement the detailed feedback from @mogaleaf !

Week3/homework/index.js
ahmthsrvsmz: okay I will.

PR #423, opened 2020-03-01
add step 3 of setup

unmeshvrije:
They have to install mysql js drivers to use MySQL with node.


README.md
```diff
@@ -67,6 +67,13 @@ mysql> grant all privileges on *.* to 'hyfuser'@'localhost';
 mysql> create database userdb;
 ```
 
+**Step 3: Installing MySQL drivers for node**
+
+We want to use MySQL with JavaScript and to this end, we use `mysqljs` driver.
```
NoerGitKat: I'd add a hyperlink to the documentation: https://github.com/mysqljs/mysql.

README.md
```diff
@@ -67,6 +67,13 @@ mysql> grant all privileges on *.* to 'hyfuser'@'localhost';
...
+* Use `node install mysql` command to install them.
```
NoerGitKat: It should be `npm install mysql`. I'd also consider asking students to install it globally, by adding the `-g` to it:

`npm install -g mysql`

README.md
```diff
@@ -67,6 +67,13 @@ mysql> grant all privileges on *.* to 'hyfuser'@'localhost';
...
+* Run `node connection-test.js` from VScode (Windows/MAC) or the terminal (Linux/MAC).
```
NoerGitKat: We can be clearer here, by mentioning that the `connection-test.js` file can be found within the directory. Students can navigate there using the CLI and then execute the command.

README.md
unmeshvrije: Added in the new commit.

README.md
unmeshvrije: Thanks for spotting this. I have added the command

README.md
unmeshvrije: Mentioned in the new commit

PR #430, opened 2020-03-18
Week 2 LESSONPLAN

unmeshvrije:


Week2/LESSONPLAN.md
```diff
@@ -0,0 +1,483 @@
...
+
+
+### 1. Key (Constraints)
+#### Explanation
+1. A column can be declared as the UNIQUE column. Such a column has UNIQUE values.
+It can also have NULL values. Thus, two rows can have same NULL value in the column that
+is declared as UNIQUE (In other words, this is a UNIQUE CONSTRAINT on that column).
+2. A column can be declared as the PRIMARY KEY column. Such a column has UNIQUE values too.
+They cannot be NULL values. Thus two rows can NEVER have same values in the column
+that is declared as PRIMARY KEY (In other words, this is a PRIMARY KEY CONSTRAINT on that column).
```
mogaleaf: I would have something about the Sequence and the auto-increment because it's really useful

Week2/LESSONPLAN.md
```diff
@@ -0,0 +1,483 @@
...
+#### Exercise
+```
+# Find out type T and Constraint C for each column.
+CREATE TABLE Airline_passengers(ticket_numer T C, passenger_name T C, date_of_birth T C, passport_number T C);
+
+Hint: A very young baby may not need a ticket!
+```
+
+#### Essence
+Primary key is a special case of UNIQUE key. UNIQUE key can be NULL and
```
mogaleaf: Maybe a word on INDEX that comes with PK and UNIQUE could be useful. forgetting to add an index on a column that is searchable is still a really common bug.

Week2/LESSONPLAN.md
unmeshvrije: It's an excellent suggestion! Thanks :)

Week2/LESSONPLAN.md
unmeshvrije: I tried very hard to not use the word INDEX because it can further create confusion with keys. But you do have a point. How can we describe it ? More importantly how could we demonstrate it in the class? I looked for some ways to populate database with lot of records and then see query performance difference with INDEXed columns and not INDEXed ones. But it is time consuming.

unmeshvrije:
> The lesson plan and the text is really clear to me.
> I have just added two points that seems essential for me.
> 
> I would just highlight that the lesson is huge compare to week 1: PK, Unique, Join, Aggregate, ERD. It's really a lot to process in one seance. I don't know if the first part (PK and Uniq) could be moved in week 1?

I think it is a very good idea. I will do so and update the homework of week 1 accordingly.
I had less content in week 1 because most people spend time in SQL installation in the class, during the class :-(

Week2/LESSONPLAN.md
mogaleaf: I understand the confusion. I would explain the Index as a big huge MAP that gives the address of the row really quickly compare to no index and having to check each row by row.
I have just assigned you a PR with a quick run to see the difference between index and no index. Hope it will be helpful.

mogaleaf:
> > The lesson plan and the text is really clear to me.
> > I have just added two points that seems essential for me.
> > I would just highlight that the lesson is huge compare to week 1: PK, Unique, Join, Aggregate, ERD. It's really a lot to process in one seance. I don't know if the first part (PK and Uniq) could be moved in week 1?
> 
> I think it is a very good idea. I will do so and update the homework of week 1 accordingly.
> I had less content in week 1 because most people spend time in SQL installation in the class, during the class :-(

Maybe an improvement for this install problem waste of time could be to ask them to install docker instead of mysql server and provide them with a docker-compose,yml file and the mysql configuration.
The would not have to add all the grant, create users etc and it should work the same for windows/linux/mac.
Then we ask them to install MySQL Workbench ( apparently works on any platform) and we skip the console.
I could try to do the install myself before the next class to see if it could be easier.

PR #431, opened 2020-03-20
js file to fill up big db

mogaleaf:
Here is a js file that fills up a simple db with same value in both columns but on with a PK and so Index and one without.
You can separate the insert and the query so if you want you can load it before the class but it's not that long on my computer.
the result is pretty clear :

Execution time with a primary key: 4ms
Execution time with no index: 10678ms

### DIFF
--- a/Week1/big-data.js
+++ b/Week1/big-data.js
@@ -1,0 +1,52 @@
+const util = require('util');
+const mysql = require('mysql');
+
+const connection = mysql.createConnection({
+    host: 'localhost',
+    user: 'hyfuser',
+    password: 'hyfpassword',
+    database: 'big',
+});
+
+
+const execQuery = util.promisify(connection.query.bind(connection));
+connection.connect();
+
+async function seedDatabase() {
+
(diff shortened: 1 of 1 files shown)


Week2/big-data.js
```diff
@@ -0,0 +1,59 @@
+const util = require('util');
+const mysql = require('mysql');
+
+const connection = mysql.createConnection({
+    host: 'localhost',
+    user: 'root',
```
unmeshvrije: could you change the user from user to 'hyfuser' and password to 'hyfpassword' ?

mogaleaf: yes you are right done !

mogaleaf:
I moved it to week 1.
I had open the pr to your branch, do you want me to open to master?

unmeshvrije:
I think it's better. Then I can directly merge it into master.
   Regards,
   Unmesh Joshi

mogaleaf:
done, pr is now on master branch (also removed an unused sleep method)

PR #432, opened 2020-03-21
Week3 mongo js

danlaudk:


danlaudk:
What if each pair looks only at one file? Then come back to discuss

Op ma 23 mrt. 2020 15:58 schreef Noer <notifications@github.com>:

danlaudk:
awaiting Noer's assessment of whether to switch over to a schema-focused rather than syntax-focused mongo lesson. Slack conversation

danlaunl:
superceded by Mongo weeks 3,4 in May 2022. can probably be closed/deleted (i think i dont have permissions)

robvk:
You are right, thanks for the effort @danlaunl !

PR #446, opened 2020-10-13
suggested change in the connection-test.js file

FatmaHameed:
Hi in HYF. I suggest adding the port number in the connection-test.js file. It could be optional in case the first code did not work for students as what happened to me. I always got access denied error before adding it. I hope it helps


Week1/connection-test.js
```diff
@@ -3,7 +3,9 @@ var connection = mysql.createConnection({
   host     : 'localhost',
   user     : 'hyfuser',
   password : 'hyfpassword',
-  database : 'userdb'
+  database : 'userdb',
+  port : 3307 // this line was not added before to the file. I added according to my selected port. It could be changed according to your selected port
```
unmeshvrije, line 7: The default port is 3306 and by default used by [mysqljs](https://github.com/mysqljs/mysql/blob/3430c513d6b0ca279fb7c79b210c9301e9315658/Readme.md)

I would suggest adding the port option as the comment and writing something like
`// port : xxxx // Uncomment this line and replace xxxx with the selected port number if you are not using default 3306`

unmeshvrije (changes requested):
Please remove package.json from the commit.
And make the port number change optional.

FatmaHameed:
Thank you @unmeshvrije for considering this. I changed the addition as per your request

FatmaHameed:
Hi @unmeshvrije , After trying to create the databases' dumbs, I realized that adding the port number is not a sustainable solution. I will try to explain how to fix the issues permanently and push it for future students because I do not want anyone to go through the same problem and waste time solving it as I did. Special thanks to Serafima because she spent a long time with me to solve it.

unmeshvrije (approved):

PR #451, opened 2020-11-01
Adds a missing rule for 1NF in Week3 README

cgduncan7:
One important rule for 1NF was missing. Added this to the quick explanation that was already provided for 1NF


unmeshvrije (approved):
LGTM

PR #453, opened 2020-12-16
Fix world.sql broken characters

tomasvanrijsse:
Tarek Almanabri fixed this file and it makes sense to share it.


tomasvanrijsse:
On github.com you can't see the difference 😩 
This is what it looked like before:
image

PR #454, opened 2020-12-20
Improve week 2

tomasvanrijsse:
I've added meaningful commit titles to explain what had to be improved

### DIFF
--- a/Week2/MAKEME.md
+++ b/Week2/MAKEME.md
@@ -26,23 +26,23 @@ This week we'll practice some more with writing SQL queries using JavaScript. Fo
 
 **Exercise 1: Keys**
 
-1. Create a table, called `Authors`. Give it the following fields: `(author_no(Primary Key), author_name, university, date_of_birth, h_index, gender)`
+1. Create a table, called `authors`. Give it the following fields: `(author_no(Primary Key), author_name, university, date_of_birth, h_index, gender)`
 2. Write a query that adds a `foreign key` column to `Authors` table that references the column `author_no`.
 Call this column `Collaborator`.
 
 **Exercise 2: Relationships**
 
-1. Create another table, called `Research_Papers` with the following fields: `(paper_id, paper_title, conference, publish_date, ...)`
-2. What is the relationship between Authors and Research papers ? Make necessary changes to `Authors` and
-`Research_Papers` tables and add more tables if necessary.
+1. Create another table, called `research_Papers` with the following fields: `(paper_id, paper_title, conference, publish_date, ...)`
+2. What is the relationship between Authors and Research papers ? Make necessary changes to `authors` and
+`research_Papers` tables and add more tables if necessary.
 3. Read exercises 3 and 4 and then add information (insert rows) of 15 authors and 30 research papers such that
 all queries in the exercises  3 and 4 will return some answers
 
 **Exercise 3: Joins**
 
-1. Write a query that prints names of all `Authors` and their corresponding `Collaborators`.
-2. Write a query that prints all columns of `Authors` and their pubished `paper_title`.
-If there is an author without any `Research_Papers`, print the information of that `Author` too.
+1. Write a query that prints names of all `authors` and their corresponding `collaborators`.
+2. Write a query that prints all columns of `authors` and their published `paper_title`.
+If there is an author without any `research_Papers`, print the information of that `author` too.
 
 **Exercise 4: Aggregate Functions**
 
--- a/Week2/MAKEME.md
+++ b/Week2/MAKEME.md
@@ -27,8 +27,8 @@ This week we'll practice some more with writing SQL queries using JavaScript. Fo
 **Exercise 1: Keys**
 
 1. Create a table, called `authors`. Give it the following fields: `(author_no(Primary Key), author_name, university, date_of_birth, h_index, gender)`
-2. Write a query that adds a `foreign key` column to `Authors` table that references the column `author_no`.
-Call this column `Collaborator`.
+2. Write a query that adds a column called `collaborator` to `authors` table that references the column `author_no`.
+For integrity add a `foreign key` on this column.
 
 **Exercise 2: Relationships**
 
--- a/Week2/README.md
+++ b/Week2/README.md
@@ -251,7 +251,7 @@ The `GROUP BY` statement groups rows that have the same values into summary rows
 The `GROUP BY` statement is often used with aggregate functions to group the result-set by one or more columns.
 
 ```sql
-SELECT COUNT(techer_number) AS no_teachers, techer_number AS teacher_number
+SELECT COUNT(teacher_number) AS no_teachers, teacher_number AS teacher_number
 FROM students
 GROUP BY teacher_number
 ```
@@ -263,7 +263,7 @@ The `Having` clause is like a `WHERE` statement but applied after the grouping h
 For example in below example, we just retrieve the teachers who teach more that three students.
 
 ```sql
-SELECT COUNT(techer_number) AS no_teachers, techer_number AS teacher_number
+SELECT COUNT(teacher_number) AS no_teachers, teacher_number AS teacher_number
 FROM students
 GROUP BY teacher_number
 HAVING COUNT(teacher_number) > 3
--- a/Week2/README.md
+++ b/Week2/README.md
@@ -204,7 +204,8 @@ FROM students
 LEFT JOIN teachers ON students.teacher_number = teachers.number
 ```
 
-It can be reversed. The reverse way of querying is called RIGHT JOIN.
+If you tried to keep all the teachers and wanted to see teachers with or without students you need a right join.
+The right join will keep all records from the joined table and discard records, from the preceding tables, that can't be joined
 
 Check out the following to get a more visual idea of what `joins` are:
 
--- a/Week2/README.md
+++ b/Week2/README.md
@@ -59,12 +59,6 @@ If you already have the table, and you just want to change a column to Primary K
 ALTER TABLE teachers ADD PRIMARY KEY (teacher_number);
 ```
 
-or you can define a primary key later:
-
-```sql
-ALTER TABLE teachers ADD CONSTRAINT PK_Person PRIMARY KEY (ID,LastName);
-```
-
 ### Foreign Key
 
 A FOREIGN KEY is a key used to link two tables together. This KEY is a field
--- a/Week2/README.md
+++ b/Week2/README.md
@@ -188,7 +188,8 @@ To learn more about this topic, check out the following:
 
 #### Inner Joins
 
-Let’s say we wanted to get a list of those students and the details of their teacher. This would be a perfect fit for an inner join, since an inner join returns records at the intersection of the two tables.
+Let’s say we wanted to get a list of students and the details of their teacher, dismissing students without a teacher.
+This would be a perfect fit for an inner join, since an inner join returns records at the intersection of the two tables.
 
 ```sql
 SELECT students.first_name, students.last_name, students.gender, students.grade, teachers.full_name
@@ -197,15 +198,21 @@ INNER JOIN teachers
 ON students.teacher_number = teachers.teacher_number
 ```
 
+The keyword `INNER` is optional and writing `JOIN teachers` will result in an inner join.
+
+#### Aliases
+
 When you join two tables there can be columns in both tables with the same name.
 To be explicit about the column you need you have to prefix the column with the table name like `table.column`
-If you want to type less you can create an alias for table names using `table as alias` or `table alias` leaving out `AS`
+If you want to type less you can create an alias for table names using `table AS alias`
 
 ```sql
 SELECT teachers.*, tq.title
 FROM teachers
-JOIN teacher_qualifications as `tq` ON tq.teacher_id = teacher.id
+JOIN teacher_qualifications AS `tq` ON tq.teacher_id = teacher.id
 ```
+
+The `AS` keyword is optional, so you could write it also like `teacher_qualifications tq`
 
 #### Right and Left Joins
 
@@ -224,6 +231,9 @@ Check out the following to get a more visual idea of what `joins` are:
 
 - [Understand SQL Joins in 10 Minutes](https://www.youtube.com/watch?v=tvMGoxmQzgQ)
 - [Joins in SQL](https://www.youtube.com/watch?v=efpFCd8iFAQ)
+
+Other types of joins are less common but can sometimes be needed.
+If you're interested you can read more about [self joins](https://www.w3schools.com/sql/sql_join_self.asp) and [full outer joins](https://www.w3schools.com/sql/sql_join_full.asp).
 
 ### Aggregate Functions in SQL
 
--- a/Week2/README.md
+++ b/Week2/README.md
@@ -246,9 +246,9 @@ FROM students
 
 ### Group By
 
-The **GROUP BY** statement groups rows that have the same values into summary rows, like "find the number of students for each teacher".
+The `GROUP BY` statement groups rows that have the same values into summary rows, like "find the number of students for each teacher".
 
-The **GROUP BY** statement is often used with aggregate functions to group the result-set by one or more columns.
+The `GROUP BY` statement is often used with aggregate functions to group the result-set by one or more columns.
 
 ```sql
 SELECT COUNT(techer_number) AS no_teachers, techer_number AS teacher_number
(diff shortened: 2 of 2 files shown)


Week2/MAKEME.md
```diff
@@ -26,23 +26,23 @@ This week we'll practice some more with writing SQL queries using JavaScript. Fo
 
 **Exercise 1: Keys**
 
-1. Create a table, called `Authors`. Give it the following fields: `(author_no(Primary Key), author_name, university, date_of_birth, h_index, gender)`
-2. Write a query that adds a `foreign key` column to `Authors` table that references the column `author_no`.
-Call this column `Collaborator`.
+1. Create a table, called `authors`. Give it the following fields: `(author_no(Primary Key), author_name, university, date_of_birth, h_index, gender)`
+2. Write a query that adds a column called `collaborator` to `authors` table that references the column `author_no`.
+For integrity add a `foreign key` on this column.
```
unmeshvrije, line 31: They are supposed to deduce that.

Week2/README.md
```diff
@@ -45,10 +45,10 @@ To define a Primary Key while creating the table, you should determine the attri
 ```sql
 CREATE TABLE teachers (
       teacher_number INT,
-      teacher_name VARCHAR(50),
+      name VARCHAR(50),
       date_of_birth DATE,
       subject TEXT,
-      gender ENUM('m', 'f'),
+      email VARCHAR(200),
```
unmeshvrije, line 51: am not sure why ENUM type needs to removed ?
It was added so that they know and practice different data types in columns

Week2/README.md
```diff
@@ -59,12 +59,6 @@ If you already have the table, and you just want to change a column to Primary K
 ALTER TABLE teachers ADD PRIMARY KEY (teacher_number);
 ```
 
-or you can define a primary key later:
-
-```sql
-ALTER TABLE teachers ADD CONSTRAINT PK_Person PRIMARY KEY (ID,LastName);
```
unmeshvrije, line 65: this way of adding primary key is syntactically different that the previous one.  So I think it should be there for reference.

Week2/README.md
```diff
@@ -75,34 +69,53 @@ To define a Foreign Key while creating the table, you can use the below query:
...
+    teacher_id int,
+    email VARCHAR(200),
     PRIMARY KEY (student_number),
     CONSTRAINT FK_TEACHER FOREIGN KEY (teacher_id) REFERENCES teachers(teacher_number)
 );
 ```
 
+A foreign key does two useful things;
+- It will verify if the related record exists, so you can't insert a row referencing a non-existing relation.
+- It will create an index on this column, giving faster results when querying on the particular column.
```
unmeshvrije, line 82: This is nice 👍

Week2/README.md
```diff
@@ -75,34 +69,53 @@ To define a Foreign Key while creating the table, you can use the below query:
...
 ```sql
 ALTER TABLE students
     ADD CONSTRAINT FK_TEACHER FOREIGN KEY (teacher_id) REFERENCES teachers(teacher_number);
 ```
 
+### Unique key
+
+The unique key is quite similar to a primary key, they both serve to check the uniqueness of a column value.
+The difference is that there can be only a single primary key, to define the record, and multiple unique keys, to define unique values.
+Foreign keys can only reference primary keys and not unique keys
```
unmeshvrije, line 95: Nice 👍

Week2/README.md
```diff
@@ -75,34 +69,53 @@ To define a Foreign Key while creating the table, you can use the below query:
...
-    student_name VARCHAR(50),
-    gender ENUM('m', 'f'),
-    PRIMARY KEY (student_number, student_name));
+    name VARCHAR(50),
+    school_id int,
+    PRIMARY KEY (student_number, school_id))
 ```
 
+Although composite keys show up in theoretical examples it isn't common to use them in practice.
+Most frameworks will add an `id` column or a prefixed id column like `student_id`
```
unmeshvrije, line 117: This is a good note !

Week2/README.md
```diff
@@ -178,34 +191,34 @@ To learn more about this topic, check out the following:
 Let’s say we wanted to get a list of those students and the details of their teacher. This would be a perfect fit for an inner join, since an inner join returns records at the intersection of the two tables.
 
 ```sql
-SELECT s.first_name, s.last_name, s.gender, s.grade, t.full_name
-FROM students s
-INNER JOIN teachers t
-ON s.teacher_number = t.teacher_number
+SELECT students.first_name, students.last_name, students.gender, students.grade, teachers.full_name
```
unmeshvrije, line 194: It is a bit verbose.
In the previous query, students know how to use aliases to improve (cut-short) the length of queries

Week2/README.md
```diff
@@ -178,34 +191,34 @@ To learn more about this topic, check out the following:
...
+FROM students
+INNER JOIN teachers
+ON students.teacher_number = teachers.teacher_number
 ```
 
-#### Self Joins
-
-A self join is a join in which a table is joined with itself (which is also called Unary relationships), especially when the table has a FOREIGN KEY which references its own PRIMARY KEY. To join a table itself means that each row of the table is combined with itself and with every other row of the table. In department, we want to get the employee's information with their direct manager. Here, each row in employees has a foreign key to itself as manager_id:
+When you join two tables there can be columns in both tables with the same name.
```
unmeshvrije, line 200: IMHO, This description is for joins in general. Not for self joins.
Perhaps, go through the example of self-join again ?

Week2/README.md
```diff
@@ -178,34 +191,34 @@ To learn more about this topic, check out the following:
...
+To be explicit about the column you need you have to prefix the column with the table name like `table.column`
+If you want to type less you can create an alias for table names using `table as alias` or `table alias` leaving out `AS`
 
 ```sql
-SELECT a.full_name AS full_name, b.full_name AS manager_name
-FROM employee a, employee b
-WHERE a.manager_id = b.id;
+SELECT teachers.*, tq.title
+FROM teachers
```
unmeshvrije, line 206: this example joins teachers table to teacher_qualifications (which is NOT a self-join).
In the previous example, `employee` is joined with `employee` (self join)

Week2/README.md
```diff
@@ -216,44 +229,44 @@ Check out the following to get a more visual idea of what `joins` are:
...
+The `GROUP BY` statement groups rows that have the same values into summary rows, like "find the number of students for each teacher".
 
-The **GROUP BY** statement is often used with aggregate functions to group the result-set by one or more columns.
+The `GROUP BY` statement is often used with aggregate functions to group the result-set by one or more columns.
 
 ```sql
-SELECT Count(s.techer_number) AS no_teachers, s.techer_number AS teacher_number
-FROM students s
-GROUP BY s.teacher_number
+SELECT COUNT(techer_number) AS no_teachers, techer_number AS teacher_number
```
unmeshvrije, line 254: Nitpick: spelling mistake from earlier query is repeated. techer=> teacher

unmeshvrije (changes requested):
I left some comments. Let's discuss them before we merge.

Week2/README.md
tomasvanrijsse: Not any particular reason except that the field didn't seem to fulfill any purpose in the subjects of this week.
I added the `email` field because it's a good example for the `UNIQUE` constraint and the `gender` field didn't seem important.
But now that we are talking about it; I would never suggest a student to use `ENUM`. It's a bad practice IMHO. Having a separate table with lookup values is almost always better because you can query those values and edit those values easily.

Week2/README.md
tomasvanrijsse: You are correct that its (slightly) syntactically different but I propose to focus on concepts rather then different syntactical operations that do exactly the same thing.

Week2/README.md
tomasvanrijsse: I made it verbose on purpose to demonstrate the use of aliases right after this example. I didn't see any explanation on aliases and heard questions about it today.

Week2/README.md
tomasvanrijsse: I removed the whole self join part and added this new line to the text about joins. 
So you are absolutely correct ;)
I have never used `SELECT * FROM table1, table2` ever and didn't find sources that advised this kind of SQL.
And even it where correct SQL it seems like a really unique thing only to be used in rare circumstances.

Week2/README.md
tomasvanrijsse: as in the previous reply on your comment; I deleted self joins and elaborated some more on regular joins.

Week2/README.md
tomasvanrijsse: Good catch. I fixed the typo.

Week2/MAKEME.md
tomasvanrijsse: I'm not a fan of the deduction-type-exercises and some students didn't know what to make of it. 
But I do concur that it's not an improvement to make it this easy.
I think the sweet spot is to explain the problem case and let them come up with the solution. 

Something like 
> 2. Authors usually help out other authors with their papers. In that role they are called a collaborator. 
Add a column to `authors` to store the value that references another author and add a key for integrity validation.

What do you think about this version?

unmeshvrije: I think `add a key for integrity validation` is extra information because I think you are meaning referential integrity here which is implied in the following phase `... value that references another author`.

Week2/README.md
unmeshvrije: The purpose was to introduce possible data types in SQL.  They can learn with experience which are good and bad (VARCHAR, TEXT etc. would also have similar arguments for and against)

Week2/README.md
unmeshvrije: It is just one extra line which IMHO does not harm any conceptual explanation.

Week2/README.md
unmeshvrije: Perhaps we could add explanation about aliases here (And in lesson plans if it is not there). PR for that would be super-welcome.

Week2/README.md
unmeshvrije: I agree that it is rare but sometimes it is the only solution for the problem.
And we do cover all other types of joins, so I would add this for completeness.

[Example of self join ](https://www.w3schools.com/sql/sql_join_self.asp) on w3schools is one example of a self-join.
Another example is `finding employees and their reporting managers who are also employees`.

Explanation of joins is good and we can keep it but I think it helps to give students a perspective on self join.

Week2/README.md
robvk: It has been 15 years since I did any SQL like this so I may be wrong, but if it is only syntactically different then I would expect the two lines to at least make the same change to the database. Am I correct that these two lines, while doing a similar thing to the database in concept, do not make the exact same change to the database?

Week2/README.md
robvk: One of the things I want to start doing more often is highlight the parts that students will be using a lot and should focus on and add things that we feel they should know about but won't be using too often to an 'extra reading' part. This is because I feel both of you are right, students get a huge amount of info dumped on them so you want to limit it to things they will need during their first year of their jobs, but you also want them to get the full picture.

So information like this where this type of self join is apparently not used a lot would be great to give them. That way they can really practice the most used cases and keep this information in the back of their mind.

Can we do something like that maybe?

unmeshvrije:
The ADD CONSTRAINT syntax may add a readable constraint name instead of an
internally generated constraint name.
   Regards,
   Unmesh Joshi

unmeshvrije:
If students complained about the amount of stuff in the database module,
then we should probably cut-down on some topics, say Normal forms or
transactions etc.
There are already too many topics we do not cover (triggers, stored
procedures etc.), So it is hard to decide which topics to keep and which
ones to remove.

Generally my viewpoint is if we are covering a concept, say join, then we
should cover all sub-concepts/all types of it.

To give an example about Javascript ("Declaring variables" topic), one
could argue that removing the keyword `var` from the curriculum is okay
because most people professionally use `const` or `let`.
But we do teach `var` because it completes the topic of variable
declaration, and because we can teach differences when it comes to hoisting
for example.

a joke:
Many of our students,  when they start working professionally, do not work
on Databases at all. So we could potentially remove the module ;-) and keep
it optional for students who want to pursue backend/full-stack jobs.


   Regards,
   Unmesh Joshi

Week2/README.md
tomasvanrijsse: On line 200 of this same file is my addition to explain aliases ;)

Week2/README.md
tomasvanrijsse: I was about to restore the ENUM but then it occurred to me; that is content for week 1, right?
https://github.com/HackYourFuture/databases/tree/master/Week1#6-what-are-data-types-as-applied-to-databases
And because it doesn't serve of explaining week 2 concepts I could remove it.

Week2/MAKEME.md
tomasvanrijsse: > Authors usually help out other authors with their papers. In that role they are called a collaborator.
Add a column to `authors` to store the `author_no` from the collaborator and add a key for integrity validation.

Is it better like this?

Week2/README.md
tomasvanrijsse: The only difference is that the removed example adds a named constraint. Which is specifically useful in SQL Server, Oracle and MS Access. Because in those DB's you have to remove the PK by name. 

I think we should stick with teaching concepts rather than syntactical oddities

Week2/README.md
tomasvanrijsse: I've added a note about extra curriculum content on line 235.

And the w3schools example of self joins is strange to me. I would rather do 
`SELECT * FROM employees JOIN employees AS managers ON managers.id = employees.manager_id`
compared to 
`SELECT * FROM employees, employees AS managers WHERE managers.id = employees.id` 

I would like to teach students to `JOIN` using `ON` and never use `WHERE` statements to join.
So teaching them this style of self joins seems counter productive.

unmeshvrije: I agree with you on stressing the use of `JOIN` keyword.

Week2/MAKEME.md
unmeshvrije: Why not remove `add a key for integrity validation` ?
or just explicitly say `add a foreign key` ?

Week2/README.md
unmeshvrije: You are right. I won't fight about it ;)
ENUM can be removed from this example.

Week2/README.md
unmeshvrije: I still disagree, but would like to move on. As long as students understand the concept, I am happy.
Let's remove it.

Week2/README.md
```diff
@@ -175,85 +188,95 @@ To learn more about this topic, check out the following:
...
 ```
 
-#### Self Joins
+The keyword `INNER` is optional and writing `JOIN teachers` will result in an inner join.
 
-A self join is a join in which a table is joined with itself (which is also called Unary relationships), especially when the table has a FOREIGN KEY which references its own PRIMARY KEY. To join a table itself means that each row of the table is combined with itself and with every other row of the table. In department, we want to get the employee's information with their direct manager. Here, each row in employees has a foreign key to itself as manager_id:
+#### Aliases
+
+When you join two tables there can be columns in both tables with the same name.
+To be explicit about the column you need you have to prefix the column with the table name like `table.column`
```
unmeshvrije, line 206: I like this explanation. I missed this is earlier review (please also add it in the LESSONPLAN.md if you can)

Week2/README.md
```diff
@@ -175,85 +188,95 @@ To learn more about this topic, check out the following:
...
+If you tried to keep all the teachers and wanted to see teachers with or without students you need a right join.
+The right join will keep all records from the joined table and discard records, from the preceding tables, that can't be joined
 
 Check out the following to get a more visual idea of what `joins` are:
 
 - [Understand SQL Joins in 10 Minutes](https://www.youtube.com/watch?v=tvMGoxmQzgQ)
 - [Joins in SQL](https://www.youtube.com/watch?v=efpFCd8iFAQ)
 
+Other types of joins are less common but can sometimes be needed.
+If you're interested you can read more about [self joins](https://www.w3schools.com/sql/sql_join_self.asp) and [full outer joins](https://www.w3schools.com/sql/sql_join_full.asp).
```
unmeshvrije, line 236: Instead of w3school link, we can provide the employee and manger join example here.
It is a better example of self join.

unmeshvrije (changes requested):
we almost there! we should be able to merge this before Christmas 🎄

Week2/MAKEME.md
tomasvanrijsse: The first part is just to tell to add the column that should accomodate values equal to `author_no` instead of the name of the collaborator for example. And I tried to formulate the intend instead of the practical implementation (without telling "why"). 

I'll rephrase it to say "add a foreign key"

Week2/README.md
tomasvanrijsse: In my opinion self joins and full outer joins aren't subjects students need to know about. But I added it back in to comply with the request from Rob (in the thread about self joins) 
> add things that we feel they should know about but won't be using too often to an 'extra reading' part.

Keeping it in the extra curriculum part seemed like a good compromise.

unmeshvrije (approved):
LGTM

PR #458, opened 2021-01-01
normalisation explantion made practical

tomasvanrijsse:


Week3/README.md
```diff
@@ -20,77 +20,93 @@ Your teachers Igor and Unmesh have made video lectures for this week's material.
...
-The rules for first normal form are:
+The first normal form requires your attributes to stay proper, in essence coming down to this list:
 
-- Rule 1 : Single valued attributes (each column should have atomic value, no multiple values)
-- Rule 2 : Attribute domain should not change
-- Rule 3 : Unique names for attributes / columns
-- Rule 4 : Order does not matter
-- Rule 5 : Repeating columns are not allowed (ie TelephoneNumber1, TelephoneNumber2, etc.)
-
-Did you get that? Neither did I! These normal form rules are written abstractly in a hard-to-understand formal language.
```
unmeshvrije, line 44: I did. But main reason I wrote them formally is that students will end up reading technical stuff online anyway. And teachers should be able to explain what they mean.

Week3/README.md
```diff
@@ -20,77 +20,93 @@ Your teachers Igor and Unmesh have made video lectures for this week's material.
...
+| ----------- | ---------- | ------- | ------------------------------------ | ------------------ | ------------------ |
+| 123         | Pooja      | Singh   | 555-861-2025, 192-122-1111           | Hoogweg 1          |                    |
+| 456         | San        | Zhang   | (555) 403-1659 Ext. 53; 182-929-2929 | High av. 3254      | Low street 2       |
+| 789         | John       | Doe     | 555-808-9633                         | Bergweg 2          |                    |
+
+This table violates rule 1 because the column for telephone numbers has multiple values.
+The easiest way to fix this is to have two separate phone columns: one for landline and one for mobile.
+
+The second problem involves `Shipping Address 1` and `Shipping Address 2`
+The proper way to solve this type of problem would be to identify entities represented in the table and separate them into their own respective tables.
```
unmeshvrije, line 56: I like this example

Week3/README.md
```diff
@@ -20,77 +20,93 @@ Your teachers Igor and Unmesh have made video lectures for this week's material.
...
-There is only one rule for third normal form: No transitive dependency (i.e. no field should depend on non-key attributes).
+**Manufacturers**
 
-This rule is very similar to the rule for 2NF with the exception that it applies to non-key attributes. The approach for making your database comply with 3NF is also similar: split up your table. You can find a complete example [here](https://en.wikipedia.org/wiki/Third_normal_form).
+| Manufacturer | Country Code |
+| ------------ | ------------ |
+| Apple        | US           |
+| Lenovo       | CN           |
+
+Update anomalies cannot occur in these tables, unlike before because there is only one value for country name for each country code.
```
unmeshvrije, line 99: Here we should explain what an `Update anomaly` is

Week3/README.md
```diff
@@ -20,77 +20,93 @@ Your teachers Igor and Unmesh have made video lectures for this week's material.
...
 
 #### Higher normal forms
 
-Besides these three normal forms, there are other higher normal forms, for example, the Boyce-Codd normal form (3.5 NF). Not all of these normal forms are equally important. Most of the time while designing databases you should aim for the third normal form. If you still did not get what are the rules for thedifferentt normal forms, don't worry. Over time you will develop a feel for it.
+Besides these three normal forms, there are other higher normal forms, for example, the Boyce-Codd normal form (3.5 NF).
+Not all of these normal forms are equally important.
+Most of the time while designing databases you should aim for the third normal form.
+If you still did not get what are the rules for the different normal forms, don't worry.
```
unmeshvrije, line 106: The lines 106 and 107 are SUPER important. Thanks for adding them.

unmeshvrije (changes requested):
Left a few comments.
Please take a look

Week3/README.md
tomasvanrijsse: I'm not going to take the credit. This was already in the readme 😊
I just added line breaks because this line was way to long

Week3/README.md
tomasvanrijsse: >> Did you get that? Neither did I!
>
> I did.

_Did you just answer your own question, again? 😛_

tomasvanrijsse: At first I didn't feel the need to rewrite the 1NF rules but mainly wanted to rewrite 2NF and 3NF.
Those two seemed way easier to explain in common English compared to theoretical definitions.
So that's why I also rewrote 1NF to stay practical for the whole normalization topic.

I'm personally a big fan of practical understanding, that helps me getting into the essence of a subject.
And I find it also easier to let students get a feeling for a subject when I teach it more practical.
That why your argument look like this to me:
"They'll read other sources online that are confusing anyway so our explanation can be confusing too"
😉 

> And teachers should be able to explain what they mean.

I agree, but wouldn't it be great if the students would grasp it straight away?
Other online sources are also easier to understand that way because they start with a basic understanding.

tomasvanrijsse: Outside of the discussion on _formal vs practical teaching_;
I removed rules 2, 3 and 4 for brevity

>Rule 2 : Attribute domain should not change

I couldn't find any proper explanation of this rule in HYF or in the given online sources.
And this one does need some explanation if you'd ask me.

> Rule 3 : Unique names for attributes / columns
> Rule 4 : Order does not matter

Why bother mentioning these two?
Since most databases will "apply" these two rules out of the box. 
These two rules only seem valuable in a theoretical case.

And I did add a new rule "Primary key (no duplicate tuples)"

unmeshvrije: > > > Did you get that? Neither did I!
> > 
> > 
> > I did.
> 
> _Did you just answer your own question, again? 😛_

Yeah, It was sarcastic.

unmeshvrije: > At first I didn't feel the need to rewrite the 1NF rules but mainly wanted to rewrite 2NF and 3NF.
> Those two seemed way easier to explain in common English compared to theoretical definitions.
> So that's why I also rewrote 1NF to stay practical for the whole normalization topic.
> 
> I'm personally a big fan of practical understanding, that helps me getting into the essence of a subject.
> And I find it also easier to let students get a feeling for a subject when I teach it more practical.
> That why your argument look like this to me:
> _"They'll read other sources online that are confusing anyway so our explanation can be confusing too"_
> 😉
> 
> > And teachers should be able to explain what they mean.
> 
> I agree, but wouldn't it be great if the students would grasp it straight away?
> Other online sources are also easier to understand that way because they start with a basic understanding.

My argument is "write what they will find in online resources (formal language) and then explain it in a way they can understand".

Week3/README.md
tomasvanrijsse: I change it to:
> Before there was a risk for inconsistencies because it was possible to update a country name in just one record.
That cannot occur anymore in these normalized tables because there is only one value for country name for each country code.

Week3/README.md
tomasvanrijsse: These rules for 1NF aren't that hard so it's not a big deal to me. 
I'll revert them back to formal language.

But when searching online there was only one source (studytonight.com) that listed the rules that where listed here.
The most common one to find is "atomic values", other sources add "prevent repeating columns" and others (like wikipedia) add "prevent duplicate tuples"

unmeshvrije: As long as we cover all correct versions of these rules, I am fine with it.

robvk: I am with Tomas here that there isn't really a widely used formal description (just the acronyms) and that different sites use different explanations. Which means I think we are free to give our own explanation of the rules. I do think all of them should be there though, even if the database checks them.

So my vote atm would be to use the new descriptions and add the 2 removed ones back.

tomasvanrijsse: After hours of googling on the topic, I hardly found any sources online that mention these rules; only on studytonight.com did. That's why I proposed to remove the three rules. (you can find sources if you search for the exact words, but you don't come across them when you search 'database normalization')

About these two: `Unique names for columns` and `Order does not matter`
I don't see the value of adding rules that can be ignored. But these two are so simple so I don't mind either to keep them in, they're easy to read and forget. So I'd coincide for sake of keeping the argument brief.

About this one: `Rule 2 : Attribute domain should not change`
This one isn't super straight forward to me and it isn't explained in our readme.
The explanation from studytonight in plain english is like this: _"This is more of a "Common Sense" rule. In each column the values stored must be of the same kind or type._
Putting that in our docs also doesn't make sense to me. The example would be "You shouldn't put dates in a varchar column".
That's common sense, even to non-IT students, right?

tomasvanrijsse: Another argument would be like the one about the exercises: https://github.com/HackYourFuture/databases/pull/460

>I think we should not focus on the exact normalization steps in the exercise.
The goal however seems legit, so I propose to keep the exercise but stress less about the normal forms.

In this case I would make the same point, to teach them "how to basically normalize" in stead of "learn all the rules"

unmeshvrije: > After hours of googling on the topic, I hardly found any sources online that mention these rules; only on studytonight.com did. That's why I proposed to remove the three rules. (you can find sources if you search for the exact words, but you don't come across them when you search 'database normalization')
> 
> About these two: `Unique names for columns` and `Order does not matter`
> I don't see the value of adding rules that can be ignored. But these two are so simple so I don't mind either to keep them in, they're easy to read and forget. So I'd coincide for sake of keeping the argument brief.
> 
> About this one: `Rule 2 : Attribute domain should not change`
> This one isn't super straight forward to me and it isn't explained in our readme.
> The explanation from studytonight in plain english is like this: _"This is more of a "Common Sense" rule. In each column the values stored must be of the same kind or type._
> Putting that in our docs also doesn't make sense to me. The example would be "You shouldn't put dates in a varchar column".
> That's common sense, even to non-IT students, right?

Common sense is not so common ;-) 
One can easily insert date in a varchar column because both support string literals. When prompting command (say `insert into`) in the class, I remember having to spell-out `INSERT`  SPACE  `INTO` 

Then again, if these rules are not that useful, we can remove them.

unmeshvrije: > Another argument would be like the one about the exercises: #460
> 
> > I think we should not focus on the exact normalization steps in the exercise.
> > The goal however seems legit, so I propose to keep the exercise but stress less about the normal forms.
> 
> In this case I would make the same point, to teach them "how to basically normalize" in stead of "learn all the rules"

I see rules as `tools to explain why some decisions were made`. If students know and understand what is a `transitive dependency` or a `partial dependency`, then they can argue about how that is violated instead of handwaving arguments (this would come handy in interviews when knowing technical terms/formal language is a plus IMHO)

In summary, I think we can simplify things not by removing rules but by adding the extra explanations where the rule are too formal/too hard to understand.

tomasvanrijsse: I didn't think of the value that it can have in interviews and debates. Perhaps because I've never had such interviews and when I interview someone I wouldn't ask such questions. 😆

That said, I have restored the rule about "consistent domain values" and also added it to the exercise
https://github.com/HackYourFuture/databases/pull/458/commits/b648aaf6082992d6cde4af0f6fa6230cd7b3869a

unmeshvrije (approved):
I think we discussed enough. Let me merge this one.
Thank you for all the changes.

PR #460, opened 2021-01-03
change focus from normalization exercises

tomasvanrijsse:
I think we should not focus on the exact normalization steps in the exercise.
The goal however seems legit, so I propose to keep the exercise but stress less about the normal forms.


unmeshvrije (approved):
I agree with changes

PR #461, opened 2021-01-10
change the week2 exercise

tomasvanrijsse:
This change is to address this issue:
https://github.com/HackYourHomework/databases-1/pull/15#discussion_r551800358

In the old situation the relation could be interpreted at 1-to-many and many-to-many.

### DIFF
--- a/Week2/MAKEME.md
+++ b/Week2/MAKEME.md
@@ -27,7 +27,7 @@ This week we'll practice some more with writing SQL queries using JavaScript. Fo
 **Exercise 1: Keys**
 
 1. Create a table, called `authors`. Give it the following fields: `(author_no(Primary Key), author_name, university, date_of_birth, h_index, gender)`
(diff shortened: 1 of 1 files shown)


unmeshvrije (changes requested):
When I wrote the homework, I did have collaborator in mind because authors of papers usually have collaborators; this can be a many-to-many relation.

So we can ask them to make a many-to-many relation table

tomasvanrijsse:
I think we where already on the same page regarding this PR.

> I wanted students to be able to write the foreign key constraint on the same table.
E.g. Employee table with manager column, author table with collaborator column etc.
In some DBs it is called "self referential integrity constraint" but the idea/implementation is exactly the same as foreign key.

That's exactly what I tried to accomplish, let them create a self referencing column.
I taught that using 'mentor' makes more sense in a one-to-many relation (which you need for this self-referencing solution).

unmeshvrije (approved):
LGTM

PR #464, opened 2021-01-13
Fix syntax error: begin transaction

tomasvanrijsse:
it's either `BEGIN` or `START TRANSACTION`


unmeshvrije (approved):
Thanks for fixing this!

PR #466, opened 2021-02-21
remove mendix video

tomasvanrijsse:
This video is talking about persisting vs non-persisting data and how to configure it using mendix software.
It's more confusing than educational.


M3kH (approved):
Maybe could be useful provide an other alternative, like an ORM demo (such as TypeORM).

unmeshvrije:
I will merge it pretending that I did all the work 😉

PR #467, opened 2021-02-21
installation instruction

tomasvanrijsse:
If students don't set this option they can't run any scripts that use the  `mysql` npm package

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,10 @@ In this module you'll learn all about a fundamental part of any software applica
 
 **Before** your first class you start you need to install the necessary software: MySQL. This differs depending on your operating system:
 
+During installation in one of the last steps you will configure the password encryption.
+Here is [a screenshot of the step](https://i.stack.imgur.com/nFnWV.jpg).
+**You must select _Legacy_ for all the given scripts to be able to connect.**
+
 - For Windows, download the [MySQL Community Server](https://dev.mysql.com/downloads/mysql/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,6 +14,10 @@ In this module you'll learn all about a fundamental part of any software applica
 
 **Before** your first class you start you need to install the necessary software: MySQL. This differs depending on your operating system:
 
+During installation in one of the last steps you will configure the password encryption.
+Here is [a screenshot of the step](https://i.stack.imgur.com/nFnWV.jpg).
+**You must select _Legacy_ for all the given scripts to be able to connect.**
```
unmeshvrije, line 19: The scripts also work with MySQL 8.0.
If they choose the _legacy_ option, then they will have to use the older version of MySQL (5.x)

I think we should promote MySQL 8.0 and make scripts work with that.

unmeshvrije (changes requested):

README.md
tomasvanrijsse: I could change the wording a little bit but this suggestion is actually meant for the v8 installation. 

There is a pull request open for two years already for the mysql npm package. It won't work with mysql v8 modern authentication. 

This is the cause that so many students couldn't get the connection test working

unmeshvrije: Ok, the new wording makes it explicit that students should still install MySQL 8.0.
I will approve it

unmeshvrije (approved):

PR #500, opened 2022-03-16
Add Q&A prep exercise Week 1

acadavid:
Hey folks,

These are the potential exercises for Q&A prep for week 1. The idea is simple: get the students to think about databases a little beyond what they just practiced during the homework. At this point, the students don't understand the concept of relationships between tables, so this exercise will be challenging and requires some outside research for those who want to solve it. The following week, they go deep into relationships; having already attempted to build relationships will better help them cement that understanding as they will connect the dots with this exercise.

The expectation is that many of them won't be able to solve it or will come up with non-suitable designs, but that's fine as we will cover a good design during the Q&A.

Lastly, we will progressively build on top of these exercises for each Q&A session, adding a little more complexity, and by the end of week 3, they should have a DB design that can reflect well a typical website like those mentioned here.

## Why are there two options?

- Option 1 refers to a recipes website. The primary motivation for this exercise is that students as a lot about when to use SQL and No SQL. We want to start building that intuition with this example.

Recipes tend to be free form, so it's a little awkward to reflect a relationship between a recipe step (cook pasta per instructions) and different recipes (which students might attempt as per the hint in the exercise). While confusing at first, the long-term goal is to show them during week 4 that a NoSQL design might be better for this particular use case. A recipe, at least the steps, looks more like a document rather than an entity, and trying to fit this into a relational DB doesn't bring great ergonomics. At the end of week 4, the teacher can then discuss the trade-offs you make with using NoSQL and SQL in a case like this.

- Option 2 refers to an IMDB-like design. At the end of week 3 the students would have a DB that reflects a similar DB to imdb.com, and we can further explore topics like indexes, normalization, etc. This DB is a good example of a relational DB, so it'd be harder to use for NoSQL comparison.

I'd love to hear your thoughts and preferences; I find both options suitable. Once we agree on one, I will change the document to reflect it.

cc @robvk @GinaSt @unmeshvrije @Breus @Tjebbee

### DIFF
--- a/Week1/QA_PREP_EXERCISE.md
+++ b/Week1/QA_PREP_EXERCISE.md
@@ -1,0 +1,17 @@
+As a preparation step for the upcoming Q&A, you will work through the following exercise. The expected result is a list of SQL `CREATE TABLE` statements that represent the tables you're being asked to design. We suggest you first think of the design of the tables (names, columns, types) and then proceed to write the SQL statements, that way you are thinking about the problem you're trying to solve (creating a database for a website), rather than the implementation first (creating a set of SQL statements).
+
+### Option 1
+
+Design the tables for a database that represents a website with food recipes. You don't need to write any specific queries at the moment, but the design should consider the need for some future queries.
+
+- Each recipe should have a name, a category, a list of ingredients, and a list of steps to follow to complete the recipe. Many recipes might share the same ingredients or the same list of steps (For example, cook pasta according to the package instructions).
+- The website will show a general page with all the recipes.
+- The website will also show recipes under a single category (Salads, Mexican, etc).
+
+### Option 2
+
+Design the tables for a database that represents a website for movies. You don't need to write any specific queries at the moment, but the design should consider the need for some future queries.
+
+- Each movie has a name, a category, a rating, a director, and a list of actors. An actor might be present in multiple movies and even be a director of others.
+- The website will show a general page with all the movies. The movies will be sorted by rating.
+- The website will show all movies under a single category (Comedy, drama, etc).
--- a/Week1/QA_PREP_EXERCISE.md
+++ b/Week1/QA_PREP_EXERCISE.md
@@ -1,17 +1,19 @@
-As a preparation step for the upcoming Q&A, you will work through the following exercise. The expected result is a list of SQL `CREATE TABLE` statements that represent the tables you're being asked to design. We suggest you first think of the design of the tables (names, columns, types) and then proceed to write the SQL statements, that way you are thinking about the problem you're trying to solve (creating a database for a website), rather than the implementation first (creating a set of SQL statements).
+As a preparation step for the upcoming Q&A, you will work through the following exercise. The expected result is a list of SQL `CREATE TABLE` statements representing the tables you will design.
 
-### Option 1
+We suggest you first think of the design of the tables (names, columns, types) and then proceed to write the SQL statements. That way, you're thinking about the problem you're trying to solve (creating a database that holds data) rather than the implementation first (creating a set of SQL statements).
 
-Design the tables for a database that represents a website with food recipes. You don't need to write any specific queries at the moment, but the design should consider the need for some future queries.
+### Exercise
 
-- Each recipe should have a name, a category, a list of ingredients, and a list of steps to follow to complete the recipe. Many recipes might share the same ingredients or the same list of steps (For example, cook pasta according to the package instructions).
-- The website will show a general page with all the recipes.
-- The website will also show recipes under a single category (Salads, Mexican, etc).
+Design the tables for a database that contains food recipes. 
 
-### Option 2
+- Each recipe should have a name, a category, a list of ingredients, and a list of steps to follow to complete the recipe. 
+  - Many recipes might share the same ingredients or the same list of steps (For example, "cook pasta according to the package instructions" could be a regular step seen in multiple recipes).
 
-Design the tables for a database that represents a website for movies. You don't need to write any specific queries at the moment, but the design should consider the need for some future queries.
+You don't need to write any specific queries now, but the design should consider that we will execute queries on the tables to extract data such as:
 
-- Each movie has a name, a category, a rating, a director, and a list of actors. An actor might be present in multiple movies and even be a director of others.
-- The website will show a general page with all the movies. The movies will be sorted by rating.
-- The website will show all movies under a single category (Comedy, drama, etc).
+- You should be able to list all the recipes.
+- You should be able to list recipes under a single category (Salads, Mexican, etc).
+
+There's a short video[1] explaining how relationships work, and you can use this for inspiration. We will expand this topic the following week.
+
+[1] https://www.youtube.com/watch?v=C3icLzBtg8I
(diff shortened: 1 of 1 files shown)


Week1/QA_PREP_EXERCISE.md
```diff
@@ -0,0 +1,17 @@
+As a preparation step for the upcoming Q&A, you will work through the following exercise. The expected result is a list of SQL `CREATE TABLE` statements that represent the tables you're being asked to design. We suggest you first think of the design of the tables (names, columns, types) and then proceed to write the SQL statements, that way you are thinking about the problem you're trying to solve (creating a database for a website), rather than the implementation first (creating a set of SQL statements).
+
```
GinaSt, line 2: I would split the last sentence in 2 sentences, for clarity.

Week1/QA_PREP_EXERCISE.md
```diff
@@ -0,0 +1,17 @@
...
+### Option 1
+
+Design the tables for a database that represents a website with food recipes. You don't need to write any specific queries at the moment, but the design should consider the need for some future queries.
+
+- Each recipe should have a name, a category, a list of ingredients, and a list of steps to follow to complete the recipe. Many recipes might share the same ingredients or the same list of steps (For example, cook pasta according to the package instructions).
+- The website will show a general page with all the recipes.
+- The website will also show recipes under a single category (Salads, Mexican, etc).
+
```
GinaSt, line 10: Maybe we can add some questions here, like:
1. Which entities can you spot in the above problem ? 
2. Which tables do you need to create to store the above data?
3. What are the relationships between those entities?

Week1/QA_PREP_EXERCISE.md
```diff
@@ -0,0 +1,17 @@
+As a preparation step for the upcoming Q&A, you will work through the following exercise. The expected result is a list of SQL `CREATE TABLE` statements that represent the tables you're being asked to design. We suggest you first think of the design of the tables (names, columns, types) and then proceed to write the SQL statements, that way you are thinking about the problem you're trying to solve (creating a database for a website), rather than the implementation first (creating a set of SQL statements).
+
+### Option 1
```
unmeshvrije, line 3: I would simply call it `Exercise 1` instead of `Option 1`. Some students might interpret it as they can do one _or_ the other.

Week1/QA_PREP_EXERCISE.md
```diff
@@ -0,0 +1,17 @@
...
+
+Design the tables for a database that represents a website with food recipes. You don't need to write any specific queries at the moment, but the design should consider the need for some future queries.
+
+- Each recipe should have a name, a category, a list of ingredients, and a list of steps to follow to complete the recipe. Many recipes might share the same ingredients or the same list of steps (For example, cook pasta according to the package instructions).
```
unmeshvrije, line 7: > For example, cook pasta according to the package instructions

This example is not clear to me. Could you please elaborate ?

Week1/QA_PREP_EXERCISE.md
```diff
@@ -0,0 +1,17 @@
...
+- The website will show a general page with all the recipes.
+- The website will also show recipes under a single category (Salads, Mexican, etc).
+
+### Option 2
+
+Design the tables for a database that represents a website for movies. You don't need to write any specific queries at the moment, but the design should consider the need for some future queries.
```
unmeshvrije, line 13: IMHO, they should at least write `CREATE TABLE` queries/statements.

Week1/QA_PREP_EXERCISE.md
unmeshvrije: Out of these 3, I think the first two questions are the most relevant and answerable for week 1.

Week1/QA_PREP_EXERCISE.md
```diff
@@ -0,0 +1,17 @@
...
+
+- Each movie has a name, a category, a rating, a director, and a list of actors. An actor might be present in multiple movies and even be a director of others.
+- The website will show a general page with all the movies. The movies will be sorted by rating.
+- The website will show all movies under a single category (Comedy, drama, etc).
```
unmeshvrije, line 17: I am not sure if we should introduce the word `website` in this exercise. We could just say that we would like to maintain a movies' database

unmeshvrije (commented):
Thanks for writing up the exercise! Let's discuss the comments

Week1/QA_PREP_EXERCISE.md
acadavid: Yes, this might be controversial. What I'm trying to do here is see if students will come up with an entity "recipe_step", which would be a somewhat awkward one given the nature of recipes. This is mostly to later illustrate that an entity relationship in this case is not the best idea, and hence using a document db can be better suited.

Week1/QA_PREP_EXERCISE.md
acadavid: I'd leave the 3rd one though, this is exactly what we're trying to hint at, getting them to think about something called "relationships" even if they haven't formally studied it.

Week1/QA_PREP_EXERCISE.md
acadavid: Yes, I have already specified that here: https://github.com/HackYourFuture/databases/pull/500/files#diff-f7c6dc5181d1cbad80620cf352d08e5da13af116551af8d4ab4a4c3532c00c11R1

Week1/QA_PREP_EXERCISE.md
acadavid: Sure, that's also fine!

unmeshvrije:
I think `designing` the tables with tools like ERD design is not the right prep-exercise because the material covers diametrically opposite stuff. `CREATE TABLE` and `INSERT VALUES` statements are covered in the weeks' material and we can ask students to think about what tables should contain and produce the list of `CREATE` / `INSERT` statements.

Moreover, for week 2, we have ERD as one of the topics. We can ask them to design some things for the week 2 prep-exercise. For this week, we can let them do mistakes while making `CREATE` table statements. They will learn next week about primary and foreign keys and will have to re-design the same tables. I hope this makes sense @GinaSt , @acadavid

acadavid:
@GinaSt @unmeshvrije @robvk The final exercise is ready and added some additional clarifications plus your suggestions. Can someone approve it so we can merge it and let the students know about it before Sunday?

unmeshvrije (approved):
The changes look good to me! I am approving because we need to give students some time to read and do these exercises

GinaSt (approved):

Week1/QA_PREP_EXERCISE.md
```diff
@@ -0,0 +1,25 @@
+As a preparation step for the upcoming Q&A, you will work through the following exercise. The expected result is a list of SQL `CREATE TABLE` statements representing the tables you will design.
+
+We suggest you first think of the design of the tables (names, columns, types) and then proceed to write the SQL statements. That way, you're thinking about the problem you're trying to solve (creating a database that holds data) rather than the implementation first (creating a set of SQL statements).
+
+### Exercise
+
+Design the tables for a database that contains food recipes. 
+
+- Each recipe should have a name, a category, a list of ingredients, and a list of steps to follow to complete the recipe.
```
GinaSt, line 9: maybe a recipe could belong to one or many categories.

PR #501, opened 2022-03-18
Preparatory Exercise for W2

GinaSt:
Hi all,

I wrote down what  I had in mind for week 2.

I tried to include some practice queries for joins but I was not sure how to properly ask this, given that they will create their own mock data.

Also, I couldn't really link the exercise to next week's material. Any ideas on how to make them think already about that (injections, transactions etc)?

Finally, I included here the exercise for the ERD. 

Let me know what you think :)

### DIFF
--- a/Week2/QA_PREP_EXERCISE.md
+++ b/Week2/QA_PREP_EXERCISE.md
@@ -15,7 +15,8 @@ Last week you designed a database for storing food recipes. How can you improve 
     > All the cakes that do not need baking
     > All the vegan and Japanese recipes
 
-(Since your table data might be different from the above questions, feel free to improvise.)
+Since your table data might be different from the above questions, feel free to improvise.
+Make sure that your database is running locally and that your queries are reproducable in .js files. 
 
 
 Having learned what an ERD is, how would you design one for the database you created? The diagram must visualize:
--- a/Week2/QA_PREP_EXERCISE.md
+++ b/Week2/QA_PREP_EXERCISE.md
@@ -1,0 +1,29 @@
+As a preparation step for the upcoming Q&A, you need to work on the following exercise, which is based on the prep exercise of the previous week. 
+
+### Exercise
+
+Last week you designed a database for storing food recipes. How can you improve your tables, based on what you learned this week?
+
+- Where can you add a primary key?
+- Which tables should be linked with a foreign key?
+- What type of relationships do you see between the tables?
+
+- Can you make queries to get:
+    > All the vegetarian recipes with potatoes
+    > All the cakes that do not need baking
+    > All the vegan and Japanese recipes
+
+(Since your table data might be different from the above questions, feel free to improvise.)
+
+
+Having learned what an ERD is, how would you design one for the database you created? The diagram must visualize:
+
+- Your entities
+- Their attributes
+- the relationships between the tables
+
+You can use [Ludichart](lucidchart.com/pages/) or [draw.io](draw.io) for creating your diagrams. 
+
+
+
+
--- a/Week2/recipes.md
+++ b/Week2/recipes.md
@@ -64,6 +64,31 @@
     Add cheddar cheese
     Add the macaroni
 
+# Tamagoyaki Japanese Omelette
+
+### Categories:
+    Vegetarian
+    Japanese
+
+### Ingredients:
+    Eggs
+    Soy sauce
+    Sugar
+    Salt
+    Olive Oil
+
+### Steps:
+    Beat the eggs
+    Add soya sauce, sugar and salt
+    Add oil to a sauce pan 
+    Bring to medium heat
+    Add some mix to the sauce pan
+    Let is cook for 1'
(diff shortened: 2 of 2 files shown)


Week2/QA_PREP_EXERCISE.md
```diff
@@ -0,0 +1,29 @@
...
+- Where can you add a primary key?
+- Which tables should be linked with a foreign key?
+- What type of relationships do you see between the tables?
+
+- Can you make queries to get:
+    > All the vegetarian recipes with potatoes
+    > All the cakes that do not need baking
+    > All the vegan and Japanese recipes
+
+(Since your table data might be different from the above questions, feel free to improvise.)
```
unmeshvrije, line 16: Let's ask them to add relevant data to their tables, so that most students will have similar data and discussions can be more focussed.

unmeshvrije (commented):
Thanks for writing up @GinaSt ! I only added one suggestion.

Week2/QA_PREP_EXERCISE.md
robvk: Think that is a good idea, can we make it a requirement to have the database running locally? And that they need to have these queries in a js file?

robvk (commented):
LGTM, nicely done!

Week2/QA_PREP_EXERCISE.md
acadavid: I have a question here. Are students free to modify the schema at this point?

For example, from this query, it seems like a `requires_baking` field would be necessary to know this, or an alternative is to query and check for steps where `%bake|baking%` is found. I think giving the students the freedom to decide is nice, they will come up with their own solutions and then we guide them into the one that makes sense.

acadavid:
@GinaSt This is a great start! Thanks.

Some thoughts:

- I think we should provide them with some basic criteria to seed the DB. For example: Add the following data to the database:

2 Japanese food recipes
3 Italian food recipes
2 Baking recipes

(and let them know it's fine to just copy paste from google ofc)

That could make it easier for them to build the examples and be able to create the queries.

- Injections here are indeed tricky. There's already an injection exercise on the homework itself, so I don't think we need further explorations. 

- Transactions are not a very natural use case unless we want to do something more elaborate like: Create an entity that has a stock of foods used in recipes (a number of carrots, XYZ grams of sugar, etc) and another one with 'created recipes' and then you can make transactions there. It does start to look like a game maybe? A game where you perhaps have to solve puzzles to find recipes and be able to create them later? I must say I'm going far here haha, and this makes the whole exercise more suitable for an SQL DB. 

In any case, I think this is a good start, we can explore more complex examples after we give this a try and figure out if it makes more sense to add additional layers of complexity.

Week2/QA_PREP_EXERCISE.md
GinaSt: I was thinking that this is more like another category, like "no-bake".

GinaSt: If we are going to ask them to add relevant data, should we give them links to existing recipes? And should these be added in the first prep-exercise?

acadavid: @GinaSt How about we ask them to modify the DB at this point as per your suggestion https://github.com/HackYourFuture/databases/pull/500#discussion_r830260735. This can be good training as to how it is to "migrate" data in models, going from a single category to multiple.

And yes, we can provide them with a few recipes, and ask them to add additional ones. Would that be sensible?

GinaSt: I added the list of recipes in week 2. The idea is that in the first week they create their own data, with a focus on cakes and Japanese recipes and then in week 2 we ask them to add specific recipes which will make them alter their schema a bit. I will try to add some more recipes later today.

@acadavid Is this what you had in mind more or less?

Week2/QA_PREP_EXERCISE.md
```diff
@@ -0,0 +1,32 @@
...
+- Which tables should be linked with a foreign key?
+- What type of relationships do you see between the tables?
+
+- Can you make queries to get:
+    > All the vegetarian recipes with potatoes
+    > All the cakes that do not need baking
+    > All the vegan and Japanese recipes
+
+Since your table data might be different from the above questions, feel free to improvise.
+Make sure that your database is running locally and that your queries are reproducable in .js files.
```
GinaSt, line 19: @robvk would this be sufficient as a requirement?

robvk: Yes, think that is clear!

robvk:
How are we looking @GinaSt @unmeshvrije @acadavid? If we want the students to do the exercise we need to merge it today

GinaStavropoulou:
Is there any other feedback that I should implement? Can someone has a look at the recipes and maybe add some more?

unmeshvrije:
The recent changes look good to me. Let's merge

PR #502, opened 2022-03-21
Propose a rubric for the evaluation of students homework

acadavid:
Hi folks,

I'm proposing here a rubric to evaluate students' homework. The follow-up for this is a "perfect homework" that we will store in a private repo to make it easier for teachers to evaluate homework in the upcoming classes.

Let me know if you find this is useful, and please feel free to comment, add or remove anything based on your criteria. If we like this, I will write one for week 2 and week 3.


robvk:
Shall we use this while we correct class35's homework? We can then see the mistakes made and see how the rubric works/add to it things you see?

acadavid:
@robvk That's a great idea! Let's do that.

PR #506, opened 2022-03-29
add prep exercise for week 3

unmeshvrije:
Very short exercise but mainly added to initiate some discussion during the class.

This can be refined for the future class(es)

### DIFF
--- a/Week3/QA_PREP_EXERCISES.md
+++ b/Week3/QA_PREP_EXERCISES.md
@@ -1,0 +1,17 @@
+As a preparation step for the upcoming Q&A, you need to work on the following exercise, which is based on the prep exercise of the previous week. 
+
+### Exercise
+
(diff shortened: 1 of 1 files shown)


Week3/QA_PREP_EXERCISES.md
```diff
@@ -0,0 +1,17 @@
+As a preparation step for the upcoming Q&A, you need to work on the following exercise, which is based on the prep exercise of the previous week. 
+
+### Exercise
+
+Last week you created an ERD for the database for storing food recipes.
+How can you normalize your database based on what you learned this week?
```
robvk, line 6: What are you expecting them to do here? Answer the question or produce a database?

unmeshvrije: Mainly to answer the question , and in order to answer it, they should produce old and new database.  I will clarify in the question

PR #507, opened 2022-04-11
Fix CRLF

unmeshvrije:
I have fixed all CRLF in this file, so the changes appear to be on every line. However it is only the new line chagne (all CRLF are replaced with LF)


robvk:
Shouldn't the lines be CRLF?

PR #509, opened 2022-05-10
add lesson plan for week 4

robvk:


Week4/LESSON_PLAN.md
```diff
@@ -0,0 +1,275 @@
...
+
+If we want to sort the find query on the timestamp to find the latest bugs we can run the following query:
+
+```js
+client.db("logging").collection("bugs").find().sort({ timestamp: -1 });
+```
+
+#### Limit
+
+The above query will give you back all of the bugs which is not great, as such we can use the limit command as follows:
```
unmeshvrije, line 123: I don't agree. One could want to see all the bugs with a certain timestamp. 

`LIMIT` clause is especially useful when top-10 or top-5 elements are sought (e.g. Top-10 countries with the most population, top-5 bugs that occurred at midnight)

Week4/LESSON_PLAN.md
```diff
@@ -0,0 +1,275 @@
...
+      },
+    },
+  ]);
+```
+
+This will give back an object with the page field in the `_id` field and the number of bugs that were logged on that page is in the `count` field!
+
+### Exercise
+
+Think of some other things to sort on or calculate and write the code to do that.
```
unmeshvrije, line 158: This exercise is too vague. We could ask a concrete question like:
`Sort all bugs by pages and then by timestamp`

Week4/LESSON_PLAN.md
```diff
@@ -0,0 +1,275 @@
...
+
+MongoDB does a lot for you, the syntax is a little different than known, but the documentation is very detailed so make use of it!
+
+## 3. Pagination
+
+Using the same bugs collection, let's look at offset and cursor-based pagination using that collection.
+
+### Explanation
+
+Go to different online shops and look at their results pages to show pagination in action.
```
unmeshvrije, line 170: This is not the explanation of pagination , sorry.

Week4/LESSON_PLAN.md
```diff
@@ -0,0 +1,275 @@
...
+
+In the above code we do it with the timestamp, in other implementations an ID can be given.
+
+### Exercise
+
+Discuss the advantages and disadvantages of both approaches.
+
+### Essence
+
+Pagination needs to be done, important to note is that it only works with sorted data!
```
unmeshvrije, line 218: The essence of pagination is that it may be useful but not necessarily with `skip-limit` always
[like explained here](https://www.mongodb.com/blog/post/paging-with-the-bucket-pattern--part-1)

unmeshvrije (commented):
I know the PR is merged but please take a look for the next class

robvk:
Thanks for the review @unmeshvrije! Have updated it in a new PR: https://github.com/HackYourFuture/databases/pull/510. If you have time I would love to hear what you think of the changes.

PR #527, opened 2022-11-29
finished databases homeworks of Week1

GOZTURK7:


GOZTURK7:
week1 homeworks databases.

PR #533, opened 2023-04-26
Minor text updates in week 1 MAKEME and general databases files

Breus:
I went through all the top-level files in the databases module and week 1. Everything actually looks very clean so I was only able to do some very minor textual updates (checked all existing links and they work).


robvk (approved):
Nice, thanks!

PR #567, opened 2024-08-25
Hanna melnyk w2 databases

hanna-melnyk:


Week1/exercise_1.js
```diff
@@ -0,0 +1,76 @@
...
+    user: 'hyfuser',
+    password: 'hyfpassword',
+    multipleStatements: true
+});
+
+
+/*Connect to SQL server*/
+connection.connect(err => {
+    if (err) {
+        return console.error('Connection error: ' + err.stack);
```
shabab477, line 14: Would be nice to know what the error message says too rather than the stack trace

Week1/exercise_1.js
```diff
@@ -0,0 +1,76 @@
...
+    }
+    console.log('Connected!');
+});
+
+
+/*SQL queries*/
+const createDatabaseAndTables =
+    `DROP DATABASE IF EXISTS meetup;
```
shabab477, line 22: Proposing to move `meetup` to a const. Like `const DB_NAME = 'meetup'` and use this in the template string. This way you can avoid typos.

Week1/exercise_1.js
```diff
@@ -0,0 +1,76 @@
...
+    USE meetup;
+    
+    CREATE TABLE Invitee (
+    invitee_no INT AUTO_INCREMENT PRIMARY KEY,
+    invitee_name VARCHAR(100),
+    invited_by VARCHAR(100)
+    );
+    
+    CREATE TABLE Room (
+    room_no INT AUTO_INCREMENT PRIMARY KEY,
```
shabab477, line 33: Small pet peeve, indentations are a bit off. Your indentations on `Meeting` are different than the rest

Week1/exercise_1.js
```diff
@@ -0,0 +1,76 @@
...
+    INSERT INTO Meeting (meeting_title, starting_time, ending_time, room_no) VALUES
+    ('Literary Classics Discussion', '2024-08-01 09:00:00', '2024-08-01 10:00:00', 1),
+    ('American Literature Seminar', '2024-08-02 11:00:00', '2024-08-02 12:00:00', 2),
+    ('Japanese Fiction Workshop', '2024-08-03 14:00:00', '2024-08-03 15:00:00', 3),
+    ('Victorian Literature Symposium', '2024-08-04 16:00:00', '2024-08-04 17:00:00', 4),
+    ('Scientific Innovations Forum', '2024-08-05 13:00:00', '2024-08-05 14:00:00', 5);`;
+
+
+// Execute the queries
+connection.query(createDatabaseAndTables, (error, results, fields) => {
```
shabab477, line 70:
```suggestion
connection.query(createDatabaseAndTables, (error, _, __) => {
```

Since you are not using the two params

shabab477 (changes requested):

Week1/exercise_2.js
```diff
@@ -0,0 +1,44 @@
...
+    host: 'localhost',
+    user: 'hyfuser',
+    password: 'hyfpassword',
+    database: 'world'
+});
+
+/*Connect to SQL server*/
+connection.connect(err => {
+    if (err) {
+        return console.error('Connection error: ' + err.stack);
```
shabab477, line 13: Same comment as this: https://github.com/HackYourFuture/databases/pull/567/files#r1736012412

shabab477, line 13: Same comment as this: https://github.com/HackYourFuture/databases/pull/567/files#r1736012412

Week1/exercise_2.js
```diff
@@ -0,0 +1,44 @@
...
+    "SELECT Name FROM city ORDER BY Population DESC LIMIT 10;",
+    "SELECT SUM(Population) AS WorldPopulation FROM country;"
+];
+
+
+// Execute each query
+queries.forEach((query, index) => {
+    connection.query(query, (error, results) => {
+        if (error) throw error;
+        console.log(`Query ${index + 1}:`);
```
shabab477, line 38:
```suggestion
        console.log(`Query ${index + 1}: ${queries[index]}`);

```

Week2/connection_query.js
```diff
@@ -0,0 +1,29 @@
+//C:\Users\knowl\Documents\hyf\databases\Week2\connection_query.js
+import mysql from 'mysql';
+
+export const createNewConnection = () => {
```
shabab477, line 4: I like how you are thinking here, but please keep in mind everytime you call this function gives an opportunity to create a new connection. Meaning you can easily overwhelm the MySQL server in a real use case scenario. You can allow only one object of connection to be passed around, this is called [Singleton pattern](https://en.wikipedia.org/wiki/Singleton_pattern) or a fixed number of connection objects to be passed around, this is called [Connection pooling](https://en.wikipedia.org/wiki/Connection_pool). 

But for the purposes of this exercise its okay 😄

Week2/exercise_1.js
```diff
@@ -0,0 +1,99 @@
...
+                return;
+            }
+
+            const columnExists = results[0].columnExists;
+
+            if (columnExists) {
+                console.log('Mentor column already exists. No changes made.');
+                resolve();
+            } else {
+                const addMentorColumnQuery = `
```
shabab477, line 51: Was adding foreign key column conditionally part of the assignment? Because you could have created the foreign key column while creating the table, removing all the code related to conditions. Could've been simpler

Week2/exercise_4.js
```diff
@@ -0,0 +1,186 @@
...
+        });
+    });
+};
+
+
+const getSumOfPapersByFemaleAuthors = (connection)  => {
+    return new Promise((resolve, reject) => {
+        const query = `
+            SELECT 
+                SUM(IF(authors.gender = 'Female', 1, 0)) AS female_author_paper_count
```
shabab477, line 39: Would prefer if you just used a `WHERE` clause

shabab477 (commented):
I have a few more files to review

Week3/assignment/transactions-create-tables.js
```diff
@@ -0,0 +1,50 @@
...
+
+// Schema for the tables
+const accountTable = {
+    account_number: 'INT PRIMARY KEY',
+    balance: 'DECIMAL(10, 2) NOT NULL'
+};
+
+const accountChangesTable = {
+    change_number: 'INT AUTO_INCREMENT PRIMARY KEY',
+    account_number: 'INT NOT NULL',
```
shabab477, line 12: I would rename this to `from_account_number` and add a column named `to_account_number`

Week3/assignment/transaction.js
```diff
@@ -0,0 +1,97 @@
+import {createNewConnection, useDatabase} from "./connection_query.js";
+
+const connection = createNewConnection();
+
+const transferAmount = (fromAccount, toAccount, amount) => {
```
shabab477, line 5: Few things to consider:
- Don't you want to check if the `fromAccount` actually contains the money available for the transfer?
- What if they account does not even exist? Do you not want to just rollback then instead of storing it in `account_changes`?

Bonus:
- You can wrap query functions that you use here to use promises, that way you  can use `async-awaits` and simplifies the code a lot.

